PR #13, opened 2015-04-28
allow defaultResponseHeaders to be specified

Andersmholmgren:


nex3:
I think a middleware is a better place for this. It would work as well, and be reusable across multiple backends.

Andersmholmgren:
This is not possible to do with middleware. Shelf io wraps HttpServer and HttpServer has defaultResponseHeaders that you cannot prevent it adding if you are using shelf today. This is really annoying for some applications where you don't want things like X-FRAME-OPTIONs etc.

Please reopen

nex3:
If you want a custom-configured `HttpServer`, that's what `serveRequests` is for. It doesn't make sense to re-expose every configurable aspect of the server through the shelf API.

Andersmholmgren:
ah I hadn't noticed `serveRequests`. Fair point. I'll use that

Issue #44, opened 2015-06-06
Add support for multi-valued headers.


nex3:
Some solution is needed for headers that have multiple values. Most headers can be losslessly merged together or split apart, and Shelf should help with the splitting when users want to access them as multiple values; some cookie-related headers can't be losslessly combined, and can only be accessed through multi-headers.

Something as simple as `Map<String, List<String>> get Message.multiHeaders` is probably enough here. It could even lazily parse the values from the underlying data.

isoos:
What's the preferred way of setting/defining multiple headers (and overriding them in `Response.change`)? (It seems to be a bit more complex than the `multiHeaders` getter.)

kevmoo:
🤷 – I don't think anyone is going to give you a good answer here.

PR #45, opened 2015-06-14
Headers values allow List<String> in addition to List.

llamadonica:
* This works because shelf uses dart:io HttpServer internally, which
  already accepts lists in its Http server.
* This to address the Set-Cookie edge-case.
* Closes dart-lang/shelf#44

### DIFF
--- a/lib/src/message.dart
+++ b/lib/src/message.dart
@@ -18,7 +18,7 @@ abstract class Message {
   /// The HTTP headers.
   ///
   /// The value is immutable.
-  final Map<String, String> headers;
+  final Map<String, Object> headers;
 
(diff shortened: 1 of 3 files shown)


nex3:
I don't think this is the right way to support multi-valued headers. Reading multi-valued headers is just as important as writing them, and this doesn't support that at all; further, the broader typing of `headers` makes it harder for people who only care about single-value headers.

The solution I'd like to see is having a separate `Map<String, List<String>> multiHeaders` map that provides access to the headers in split form. For most headers this would automatically handle splitting both headers that were provided with multiple values and those that were folded together, and it would have special handling for the cookie headers that don't support normal folding.

PR #53, opened 2015-11-16
Fixed the comparison in the content-type header to be case insensitive

Pacane:


kevmoo:
LGTM, but @nex3 should verify. Please rebase your commits into one commit.

And thanks1

kevmoo:
Also, update the commit message to include `Closes URL-to-issue` – Github does magic to link these together

Pacane:
@kevmoo There you go!

lib/src/message.dart
```diff
@@ -138,13 +138,16 @@ abstract class Message {
 Map<String, String> _adjustHeaders(
     Map<String, String> headers, Encoding encoding) {
   if (headers == null) headers = const {};
-  if (encoding == null) return headers;
-  if (headers['content-type'] == null) {
-    return addHeader(headers, 'content-type',
+
+  var caseInsensitiveHeaders = new CaseInsensitiveMap.from(headers);
```
nex3: Don't copy the headers map if `encoding` is `null`.

Also, if you just re-assign to `headers`, you can avoid wrapping the const map in the case that `headers` was null.

test/message_change_test.dart
```diff
@@ -59,14 +59,14 @@ void _testChange(Message factory(
     var request = factory(headers: {'header1': 'header value 1'});
     var copy = request.change(headers: {});
 
-    expect(copy.headers, same(request.headers));
```
nex3: These `same()` checks are important. They test that the headers map isn't copied more than necessary.

test/response_test.dart
```diff
@@ -140,4 +140,41 @@ void main() {
       expect(response.read, throwsStateError);
     });
   });
+
+  group("content type should be preserved when setting encoding", () {
```
nex3: These should by in `message_test.dart`.

nex3:
Looking pretty good! This is definitely the right direction, there are just a few changes to make.

Pacane:
There you go @nex3 Thanks for the review.

nex3:
Thanks for the fix!

PR #61, opened 2016-10-24
Accept List<int> message bodies.

nex3:
This also automatically sets the content-length header when possible,
and works around dart-lang/sdk#27660.

### DIFF
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -78,7 +78,7 @@ class Body {
 
     // Non-ASCII code units between U+0080 and U+009F produce 8-bit characters
     // with the high bit set.
-    return bytes.any((byte) => byte & 0x80 != 0);
+    return bytes.every((byte) => byte & 0x80 == 0);
   }
 
   /// Returns a [Stream] representing the body.
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -5,8 +5,8 @@
 import 'dart:async';
 import 'dart:convert';
 
-import 'package:collection/collection.dart';
 import 'package:async/async.dart';
+import 'package:collection/collection.dart';
 
 /// The body of a request or response.
 ///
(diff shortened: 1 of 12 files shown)


lib/src/body.dart
```diff
@@ -5,6 +5,7 @@
 import 'dart:async';
 import 'dart:convert';
 
+import 'package:collection/collection.dart';
```
jmesserly: not sure if you care about sorting imports, but mentioning in case you do.
(Personally I don't think it's worth worrying about)

lib/src/body.dart
```diff
@@ -18,31 +19,70 @@ class Body {
...
+
+  /// Returns whether [bytes] is plain ASCII.
+  ///
+  /// [codeUnits] is the number of code units in the original string.
+  static bool _isPlainAscii(List<int> bytes, int codeUnits) {
+    // Most non-ASCII code units will produce multiple bytes and make the text
+    // longer.
+    if (bytes.length != codeUnits) return false;
+
+    for (var byte in bytes) {
```
jmesserly: totally optional but this could also be:
```dart
return bytes.every((b) => b & 0x80 == 0);
```

... instead of the loop. I guess it depends on how low level optimized this code is trying to be. (Fastest would be a `for (var i = 0, len = bytes.length; i < len; ++i)` style loop, IIRC)

lib/src/message.dart
```diff
@@ -134,17 +143,48 @@ abstract class Message {
 ///
```
jmesserly: maybe mention in the doc comment (1 line above) that this also adds info about the content-length?

lib/src/body.dart
nex3: Done.

lib/src/body.dart
nex3: Done. I'm not *too* worried about efficiency here, since it's only hit when the user doesn't pass in an explicit encoding.

lib/src/message.dart
nex3: Done in `new Request()` and `new Response()` (the `Message` class isn't actually public).

PR #66, opened 2016-12-06
Update Transfer-Encoding logic.

### LINKED ISSUE #64: Endless get request starting 0.6.7+1 version
In my application, when a OPTIONS request is done, the server didn't answer (in the network console, the request is pending for 3 minutes).
The issue occurs on Dartium, Chrome, Safari, Firefox.

The same case is running well with previous Shelf versions.

nex3:
This should be the last CL in this unfortunate series. The adapter
requirements have been updated to consistently ensure that a Message's
body is in the chunked format if and only if its headers indicate that
it should be.

Closes #64


lib/shelf_io.dart
```diff
@@ -136,11 +141,35 @@ Future _writeResponse(Response response, HttpResponse httpResponse) {
...
   response.headers.forEach((header, value) {
     if (value == null) return;
     httpResponse.headers.set(header, value);
   });
 
+  var coding = response.headers['transfer-encoding'];
+  if (coding != null && !equalsIgnoreAsciiCase(coding, 'identity')) {
+    // If the response is already in a chunked encoding, de-chunk it because
+    // otherwise `dart:io` will try to add another layer of chunking. See
+    // sdk#27886.
```
efortuna: gross. (the issue)
Maybe add a TODO to remove this branch afterwards?

nex3: Done.

PR #69, opened 2016-12-20
Fixes for latest pkg/http_parser

kevmoo:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.6.8
+
+* Require the latest version of `package:http_parser`.
```
nex3: I don't like requiring patch versions, even if they fix relevant bugs. There's no reason the version solver should ever select 3.1.0 anymore anyway.

Also, if we did do this, we'd release it as 0.6.7+3, not 0.6.8.

kevmoo: I was hoping to help folks who are just running `pub get` – they may still get the old thing.

Also, I was planning to push a few more commits before I wanted to release – agree w/ the eventual version, though.

nex3: If users just run "pub get", they'll also get the older `shelf` version, so I don't think there's much help there.

kevmoo:
@nex3 PTAL

kevmoo:
I'm OOO. Please merge :-)

PR #75, opened 2017-06-21
Adds SSL support

keaner:
Adds serveSecure method which passes through to bindSecure


kevmoo:
I like it – although I wonder if we should just add `securityContext` as an optional argument to serve.

@nex3 ?

lib/shelf_io.dart
```diff
@@ -57,6 +57,23 @@ void serveRequests(Stream<HttpRequest> requests, Handler handler) {
...
 
+/// Starts an SSL [HttpServer] that listens on the specified [address] and
+/// [port] and sends requests to [handler].
+///
+/// See the documentation for [HttpServer.bindSecure] for more details on
+/// [address], [port], and [backlog].
+Future<HttpServer> serveSecure(
+    Handler handler, address, int port, SecurityContext serverContext,
+    {int backlog}) {
+  if (backlog == null) backlog = 0;
```
jakemac53: nit: `backlog ??= 0`

nex3:
> I like it – although I wonder if we should just add `securityContext` as an optional argument to serve.

I like this idea.

keaner:
Cool, I'll make SecurityContext an optional arg and add tests

kevmoo:
Awesome.

keaner:
Updated based on recommendations and added two tests.

I originally tried to add the certs as separate files but I was running into this error:
"Unsupported operation: Cannot extract a file path from a data URI"

lib/shelf_io.dart
```diff
@@ -30,11 +30,22 @@ export 'src/io_server.dart';
...
-  return HttpServer.bind(address, port, backlog: backlog).then((server) {
+/// If a [securityContext] is provided an HTTPS server will be started.
+////
+/// See the documentation for [HttpServer.bind] and [HttpServer.bindSecure]
+/// for more details on [address], [port], and [backlog].
+Future<HttpServer> serve(Handler handler, address, int port,
+    {SecurityContext securityContext, int backlog}) {
+  backlog ??= 0;
+  if (securityContext == null) {
+    return HttpServer.bind(address, port, backlog: backlog).then((server) {
```
nex3: Use async/await here rather than duplicating `then()` calls.

test/shelf_io_ssl_test.dart
```diff
@@ -0,0 +1,69 @@
...
+
+        _serverPort = server.port;
+      }));
+}
+
+Future<HttpClientRequest> _scheduleGet() {
+  return schedule/*<Future<HttpClientRequest>>*/(() {
+    return client.getUrl(Uri.parse('https://localhost:$_serverPort/'));
+  });
+}
```
nex3: Can you make this part of `shelf_io_test.dart`?

lib/shelf_io.dart
keaner: Done!

test/shelf_io_ssl_test.dart
keaner: Done!

PR #76, opened 2017-07-26
Run tests with IPv6 address

natebosch:


natebosch:
/sigh

Travis doesn't support IPv6, internal tests don't support IPv4 😢 

Will likely just disable the test internally.

kevmoo:
You can do a sniff of Travis and switch to ipv4 – seems reasonable.

kevmoo:
See https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

natebosch:
Good idea!

updated PTAL

natebosch:
Ugh, now the buildbots don't like using IPv6

https://luci-logdog.appspot.com/v/?s=chromium%2Fbb%2Fclient.dart.packages%2Fpackages-linux-shelf%2F1077%2F%2B%2Frecipes%2Fsteps%2Fpub_run_test%2F0%2Fstdout

test/io_server_test.dart
```diff
@@ -16,7 +16,10 @@ import 'test_util.dart';
 void main() {
   var server;
   setUp(() async {
-    server = await IOServer.bind(InternetAddress.LOOPBACK_IP_V4, 0);
+    var address = Platform.environment.containsKey('TRAVIS')
+        ? InternetAddress.LOOPBACK_IP_V4
+        : InternetAddress.LOOPBACK_IP_V6;
```
nex3: Can you add a comment explaining why we're testing on IPv6 on Travis and only Travis?

natebosch: Sure. The chances that the `git blame` becomes at all difficult to find before the comment becomes a lie seems incredibly low but we can add one for now.

https://github.com/dart-lang/shelf/pull/89

PR #79, opened 2017-08-25
Make `Handler` return a `FutureOr`

### LINKED ISSUE #78: `Handler` should return a `FutureOr<Response>`
We couldn't previously express this so we had to go with `dynamic`. Now that we have a better way to say it we should use it.

I suppose we should call this a breaking change, though really all it does is shift a runtime error to a static error.

natebosch:
Fixes #78

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ dart:
   - dev
   - stable
   - 1.22.1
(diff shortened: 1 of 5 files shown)


kevmoo:
Drop 1.20 from travis, @natebosch ?

PR #82, opened 2017-08-29
Remove HijackCallback and OnHijackCallback

### LINKED ISSUE #81: Finish deprecating HijackCallback and OnHijackCallback
Started in https://github.com/dart-lang/shelf/commit/790d193e4ef9ebe7b0ac67aa66d59723d734b64b

- The signature of `void hijack(Function)` should become `void hijack(void onHijack(StreamChannel<List<int>> channel))` since today it's too loose.
- Remove checking against the old typedef
- Other cleanup???

natebosch:
Closes #81

These have been deprecated, remove them completely.

### DIFF
--- a/lib/src/request.dart
+++ b/lib/src/request.dart
@@ -11,12 +11,9 @@ import 'hijack_exception.dart';
 import 'message.dart';
 import 'util.dart';
 
-/// A callback provided by a Shelf handler that's passed to [Request.hijack].
-typedef void _HijackCallback(StreamChannel<List<int>> stream);
-
 /// A callback provided by a Shelf adapter that's used by [Request.hijack] to
 /// provide a [HijackCallback] with a socket.
-typedef void _OnHijackCallback(_HijackCallback callback);
+typedef void _OnHijackCallback(void callback(StreamChannel<List<int>> channel));
 
 /// Represents an HTTP request to be processed by a Shelf application.
 class Request extends Message {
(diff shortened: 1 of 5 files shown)


lib/src/request.dart
```diff
@@ -13,14 +13,12 @@ import 'message.dart';
 import 'util.dart';
 
 /// A callback provided by a Shelf handler that's passed to [Request.hijack].
-@Deprecated("Will be removed in shelf 0.7.0.")
-typedef void HijackCallback(
+typedef void _HijackCallback(
```
kevmoo: Wouldn't you want to make this take a StreamChannel, too?

natebosch: I can if I change here: https://github.com/dart-lang/shelf/blob/17182d1da9a25768a66a45b87e94a65f6809f91a/lib/src/request.dart#L146

Want me to do that?

kevmoo: Seems reasonable, right? Keep 'em consistent? If we're forcing one end to use StreamChannel, the user has already dealt with the extra class/library.

Am I missing anything?

natebosch: Done

lib/src/request.dart
```diff
@@ -13,14 +12,11 @@ import 'message.dart';
 import 'util.dart';
 
 /// A callback provided by a Shelf handler that's passed to [Request.hijack].
-@Deprecated("Will be removed in shelf 0.7.0.")
-typedef void HijackCallback(
-    Stream<List<int>> stream, StreamSink<List<int>> sink);
+typedef void _HijackCallback(StreamChannel<List<int>> stream);
```
kevmoo: Delete this and use the inline syntax in `_OnHijack.run` – right?

natebosch: Yup good call. We still need _OnHijackCallback since it's used as a class field and we don't want to use the new function type syntax yet.

kevmoo:
w00t!

PR #87, opened 2017-09-14
Stop using pkg/scheduled_test

kevmoo:
Closes https://github.com/dart-lang/shelf/issues/80

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: shelf
-version: 0.7.0
+version: 0.7.1-dev
 author: Dart Team <misc@dartlang.org>
 description: Web Server Middleware for Dart
 homepage: https://github.com/dart-lang/shelf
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: shelf
-version: 0.7.0
+version: 0.7.1-dev
 author: Dart Team <misc@dartlang.org>
 description: Web Server Middleware for Dart
 homepage: https://github.com/dart-lang/shelf
 environment:
-  sdk: '>=1.22.0 <2.0.0-dev.infinity'
+  sdk: '>=1.23.0 <2.0.0-dev.infinity'
```
natebosch: too bad we don't allow a different min SDK for `test/` from `lib/`

kevmoo: Most of our ecosystem keeps current – not a huge deal

PR #89, opened 2017-09-24
Try IPv6 and switch to IPv4 if it doesn't work

natebosch:


test/io_server_test.dart
```diff
@@ -16,6 +16,7 @@ import 'test_util.dart';
 void main() {
   var server;
   setUp(() async {
+    // Travis only supports IPv4 while other test platforms only support IPv6
```
nex3: What does "other test platforms" mean? How do we know all non-Travis platforms will support IPv6? Wouldn't it be more robust to auto-detect which is supported?

natebosch: I switched it to a try/catch.

PR #91, opened 2017-10-18
Spelling and dartdoc references

kevmoo:


lib/src/server.dart
```diff
@@ -30,7 +30,7 @@ abstract class Server {
   ///
   /// Requests to this URL or any URL beneath it are handled by the handler
   /// passed to [mount]. If [mount] hasn't yet been called, the requests wait
-  /// until it is. If [close] has been called, [handler] will not be invoked;
+  /// until it is. If [close] has been called, `handler` will not be invoked;
```
nex3: I'd change "`handler`" to "the handler".

kevmoo: ack

Issue #37, opened 2015-06-05
Add access the connection info from HttpRequest


DartBot:
<a href="https://github.com/Fox32"><img src="https://avatars.githubusercontent.com/u/648527?v=3" align="left" width="96" height="96"hspace="10"></img></a> **Issue by [Fox32](https://github.com/Fox32)**
_Originally opened as dart-lang/sdk#21026_

---

shelf gives a nice abstraction above the underling communication. But it would be nice to access the connectionInfo (containg address and port of the remote host) from HttpRequest in the case that the adapter from shelf_io is used.

This is important for logging requests.

If I understand the design of the shelf Request class correctly, than the information should be added in the context map.

DartBot:
<a href="https://github.com/floitschG"><img src="https://avatars.githubusercontent.com/u/8631949?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [floitschG](https://github.com/floitschG)**

----

*Added [Area-Library](../labels/Area-Library), [Library-IO](../labels/Library-IO), [Triaged](../labels/Triaged) labels.*

DartBot:
<a href="https://github.com/skabet"><img src="https://avatars.githubusercontent.com/u/22043?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [skabet](https://github.com/skabet)**

----

*Removed [Area-Library](../labels/Area-Library), [Library-IO](../labels/Library-IO) labels.*
*Added [Area-Pkg](../labels/Area-Pkg), [Pkg-Shelf](../labels/Pkg-Shelf) labels.*

DartBot:
<a href="https://github.com/Fox32"><img src="https://avatars.githubusercontent.com/u/648527?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [Fox32](https://github.com/Fox32)**

----

I would really like to help with this enhancement after shelf is moved to github.

DartBot:
<a href="https://github.com/nex3"><img src="https://avatars.githubusercontent.com/u/188?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [nex3](https://github.com/nex3)**

----

https://github.com/dart-lang/shelf

DartBot:
[...]

PR #92, opened 2017-10-21
 Add raw dart:io HttpRequest to Request context

jhayley:
This addresses #37.

I figure this would need to be documented somewhere, but I couldn't decide where the best place for it would be, so let me know what your suggestions are!

### DIFF
--- a/lib/shelf_io.dart
+++ b/lib/shelf_io.dart
@@ -134,7 +134,7 @@ Request _fromHttpRequest(HttpRequest request) {
       headers: headers,
       body: request,
       onHijack: onHijack,
-      context: { 'io.httprequest': request });
+      context: {'io.httprequest': request});
 }
 
 Future _writeResponse(Response response, HttpResponse httpResponse) {
--- a/lib/shelf_io.dart
+++ b/lib/shelf_io.dart
@@ -134,7 +134,7 @@ Request _fromHttpRequest(HttpRequest request) {
       headers: headers,
       body: request,
       onHijack: onHijack,
-      context: {'io.httprequest': request});
+      context: {'io.httprequest.connectionInfo': request.connectionInfo});
 }
 
 Future _writeResponse(Response response, HttpResponse httpResponse) {
--- a/lib/shelf_io.dart
+++ b/lib/shelf_io.dart
@@ -14,6 +14,10 @@
 /// wire as they're received. See [`HttpResponse.bufferOutput`][bufferOutput]
 /// for more information.
 ///
+/// `Request`s passed to a `Handler` will contain the
+/// `"shelf.io.connection_info"` `Request.context` property, which holds the
+/// `HttpConnectionInfo` object from the underlying `HttpRequest`.
+///
 /// [bufferOutput]: https://api.dartlang.org/apidocs/channels/stable/dartdoc-viewer/dart:io.HttpResponse#id_bufferOutput
 import 'dart:async';
 import 'dart:io';
--- a/lib/shelf_io.dart
+++ b/lib/shelf_io.dart
@@ -133,7 +133,8 @@ Request _fromHttpRequest(HttpRequest request) {
       protocolVersion: request.protocolVersion,
       headers: headers,
       body: request,
-      onHijack: onHijack);
+      onHijack: onHijack,
+      context: { 'HttpRequest': request });
 }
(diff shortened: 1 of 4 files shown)


jhayley:
Travis is failing with `502 Bad Gateway trying to find package stream_channel at https://pub.dartlang.org.`, I assume that's unrelated

kevmoo:
@jhayley Travis seems happy now...

test/shelf_io_test.dart
```diff
@@ -455,6 +455,23 @@ void main() {
     expect(stream.hasNext, completion(isFalse));
   });
 
+  test('includes the dart:io HttpRequest in the request context', () async {
+    await _scheduleServer((request) {
+      expect(request.context.containsKey('io.httprequest'), isTrue);
+      expect(
+          request.context['io.httprequest'], new isInstanceOf<HttpRequest>());
```
nex3: You can use

```dart
expect(request.context, containsPair('io.httprequest', new isInstanceOf<HttpRequest>()));
```

to do this a little more simply.

test/shelf_io_test.dart
```diff
@@ -455,6 +455,23 @@ void main() {
...
+
+      HttpRequest httpRequest =
```
nex3: Nit: don't type-annotate local variables.

jhayley:
Alright, that makes sense! After looking it over I think those fields you mentioned are the only ones that would be relevant, so I'll get those in

jhayley:
So, in writing an ssl test for the certificate key specifically, it looks like the test certificate that's being used isn't getting passed down into the httprequest. Is there a reason this isn't happening? There is a one coming from the response though.

I could ignore this and just test for the existence of the key but that seems a little disingenuous

nex3:
> So, in writing an ssl test for the certificate entry specifically, it looks like the test certificate that's being used isn't getting passed down into the httprequest. Is there a reason this isn't happening? There is a one coming from the response though.

I'm not sure I follow. What does the test code look like, and where is it failing?

jhayley:
I've added this test in the ssl tests group in shelf_io_test:
```
    test('includes the X509Certificate in an ssl request\'s context', () async {
      await _scheduleServer((request) {
        expect(
            request.context,
            containsPair('io.httprequest.certificate',
                new isInstanceOf<X509Certificate>()));

        return syncHandler(request);
      }, securityContext: securityContext);

      var req = await _scheduleSecureGet();

      var response = await req.close();
      expect(response.statusCode, HttpStatus.OK);
    });
```

Which errors out at the `expect` because io.httprequest.certificate and therefore the request.certificate in _fromHttpRequest is null. I figured since the ssl tests are passing a security context to the client, there would be a certificate passed down, but maybe I'm missing something.

The docs for HttpRequest.certificate say that it will be null "if the connection is not a secure TLS or SSL connection, or if the server does not request a client certificate, or if the client does not provide one", so is this falling into the second case?

nex3:
Ah, I see. Just using SSL isn't enough to set `HttpRequest.certificate`; it's a client-side certificate, which means the server has to request it (by setting `requestClientCertificate: true` when calling [`HttpServer.bindSecure()`](https://api.dartlang.org/stable/1.24.2/dart-io/HttpServer/bindSecure.html)) and the client has to send it (by passing a security context to `HttpClient()`, which you're doing). We don't currently expose `requestClientCertificate` in `shelf_io.serve`... you could add that if you want, but I don't think anyone really uses client certificates so I'd be fine limiting this pull request to exposing the connection info.

lib/shelf_io.dart
```diff
@@ -133,7 +133,8 @@ Request _fromHttpRequest(HttpRequest request) {
       protocolVersion: request.protocolVersion,
       headers: headers,
       body: request,
-      onHijack: onHijack);
+      onHijack: onHijack,
+      context: {'io.httprequest.connectionInfo': request.connectionInfo});
```
nex3: I think I'd call this field `"shelf.io.connection_info"`. We want to make sure it's scoped to this package, but since everything coming from IO is inherently going to involve an HTTP request, we don't really need to add that to the hierarchy.

jhayley:
Renamed and documented!

jhayley:
Whoops, will get that dartfmt'd tonight

PR #93, opened 2017-12-20
Update `createMiddleware` arguments to use `FutureOr`.

kevmoo:
Note: this change is not breaking for the runtime behavior, but it might
cause new errors during static analysis due the the type changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## 0.8.0
+
+* Update `createMiddleware` arguments to use `FutureOr`.
+
+  * Note: this change is not breaking for the runtime behavior, but it might
+    cause new errors during static analysis due the the type changes. When
+    updating packages that depend on `shelf`, it is safe to include `0.7.1` in
+    your version range.
+
+      `shelf: '>=0.7.1 <0.9.0'`
+
 ## 0.7.1
 
 * The `shelf_io` server now adds a `shelf.io.connection_info` field to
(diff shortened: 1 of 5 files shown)


kevmoo:
@natebosch – what do you think about the version change?

I had strong mode errors in the tests without changes. But this was when we were passing in `void fail(...)` functions into the callbacks.

`dynamic xyz(...)` works fine. `void` and `Object` blow things up.

Worth going to 0.8?

natebosch:
I do think we should treat this as a breaking change. I think there are some small holes that we're closing where code would have been fine before but is a static error now.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 0.7.2
+
+* Update `createMiddleware` arguments to use `FutureOr`.
+
+  * Note: this change is not breaking for the runtime behavior, but it might
+    cause new errors during static analysis due the the type changes.
+
+* Updated minimum Dart SDK to `1.24.0`, which added `FutureOr`.
```
nex3: FYI: We generally don't list dependency range changes (including SDK constraint changes) in the changelog. Pub will handle these automatically, which means making the downstream consumer think about them adds cognitive overhead for little value.

kevmoo: ...I always thought it can help users understand why they are not getting the latest version.

nex3: I don't think most users have a workflow that involves looking at their dependencies' changelogs when trying to figure out the version solver's behavior. Once the new version solver lands, they should be able to just run `pub upgrade $package` and get good diagnostics if the latest version isn't available.

kevmoo: Can't wait!

Issue #88, opened 2017-09-17
is there a way for shelf to share the connections among isolates ?


sandrobilbeisi:
I use HttpServer.bind directly with shared: true on a system with up to four isolates
and I use the chief isolate as a data mothership

is it possible to do this whilst using shelf ?

sandrobilbeisi:
it looks to me that shelf_io.dart doesn't have the 'shared' parameter ( not even for passthrough)
https://github.com/dart-lang/shelf/blob/master/lib/shelf_io.dart

I will attempt a monkey patch and communicate my findings

sandrobilbeisi:
hmmm ... I just noticed that issue #56 references an attempted patch at
https://github.com/dart-lang/shelf/pull/56/commits/f51aebb343412b6e6011b039d16a1ffb303b3aeb

kevmoo:
Or just use `serveRequests`. `HttpServer` is `Stream<HttpRequest> requests` – so follow your tricks to spin up a multi-isolate `HttpServer` – then pass it in there...

sandrobilbeisi:
that's awesome ! thanks @kevmoo 

you forced me to read the comments and the code more thoroughly !

I'm trying it now and it seems to work ,...
 I'll post a skeleton example

kevmoo:
🤘

From mobile. Sorry for brevity and/or spelling.

KOGI:
@sandrobilbeisi did you ever get a skeleton example put together for this? It would be supremely helpful for me!

kevmoo:
See https://github.com/dart-lang/isolate/blob/master/example/http_server.dart

you can adapt this on top of shelf

KOGI:
Thank you!

PR #109, opened 2018-12-06
Update shelf_io.dart

emeyex:
Pass `shared` parameter through to `HttpServer.bind` and `HttpServer.bindSecure` (i.e., to enable port sharing across separate isolates)


emeyex:
CLA is signed!

emeyex:
Updated changelog!

emeyex:
Looks like this would resolve issue #88 (i.e., as an alternative to calling `serveRequests` directly)

PR #112, opened 2019-03-28
Return the correct HTTP status code for badly formatted requests.

isoos:
A bad request should return 400 and not 500.
Ref: https://github.com/dart-lang/pub-dartlang-dart/issues/2081

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.5.
+
+* Return the correct HTTP status code for badly formatted requests.
+
 ## 0.7.4+1
 
 * Allow `stream_channel` version 2.x
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.5.
+## 0.7.5
 
 * Return the correct HTTP status code for badly formatted requests.
 
--- a/lib/shelf_io.dart
+++ b/lib/shelf_io.dart
@@ -74,8 +74,10 @@ Future handleRequest(HttpRequest request, Handler handler) async {
   try {
     shelfRequest = _fromHttpRequest(request);
   } catch (error, stackTrace) {
-    var response =
-        _logTopLevelError('Error parsing request.\n$error', stackTrace);
+    _logTopLevelError('Error parsing request.\n$error', stackTrace);
+    final response = Response(400,
+        body: 'Bad Request',
+        headers: {HttpHeaders.contentTypeHeader: 'text/plain'});
     await _writeResponse(response, request.response);
     return;
   }
@@ -203,10 +205,11 @@ Response _logError(Request request, String message, [StackTrace stackTrace]) {
   buffer.writeln();
   buffer.write(message);
 
-  return _logTopLevelError(buffer.toString(), stackTrace);
+  _logTopLevelError(buffer.toString(), stackTrace);
+  return Response.internalServerError();
 }
 
-Response _logTopLevelError(String message, [StackTrace stackTrace]) {
+void _logTopLevelError(String message, [StackTrace stackTrace]) {
   var chain = Chain.current();
   if (stackTrace != null) {
     chain = Chain.forTrace(stackTrace);
@@ -218,5 +221,4 @@ Response _logTopLevelError(String message, [StackTrace stackTrace]) {
   stderr.writeln('ERROR - ${DateTime.now()}');
   stderr.writeln(message);
   stderr.writeln(chain);
-  return Response.internalServerError();
 }
--- a/lib/shelf_io.dart
+++ b/lib/shelf_io.dart
@@ -73,15 +73,22 @@ Future handleRequest(HttpRequest request, Handler handler) async {
   Request shelfRequest;
   try {
     shelfRequest = _fromHttpRequest(request);
+  } on ArgumentError catch (error, stackTrace) {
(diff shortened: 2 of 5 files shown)


isoos:
I've set the version to `0.7.5` because I also wanted to fix the `/https://example.com/` path parsing bug, but it may not be that straightforward, and it would be better to release `0.7.5` now and fix the other bug later. wdyt?

lib/shelf_io.dart
```diff
@@ -74,8 +74,10 @@ Future handleRequest(HttpRequest request, Handler handler) async {
   try {
     shelfRequest = _fromHttpRequest(request);
   } catch (error, stackTrace) {
```
jonasfj: Hmm, we probably have to distinguish between errors that happen because of a bug in `_fromHttpRequest` and errors that happen because the request is malformed.

isoos: Ack, updated the PR: only return 400 when URL-validation check failed, otherwise keep the 500.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.7.5.
```
kevmoo: Drop the trailing `.`

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+* Return the correct HTTP status code for badly formatted requests.
```
kevmoo: `UriArgumentError` is now part of the public API, right? Should document that here, too!

isoos: Good question. Should it become part of the public API? If we make it a private class, it will still be the same `ArgumentError` as previously, and that way it doesn't change the public API.

lib/src/request.dart
```diff
@@ -167,20 +167,20 @@ class Request extends Message {
         this.handlerPath = _computeHandlerPath(requestedUri, handlerPath, url),
         this._onHijack = onHijack,
         super(body, encoding: encoding, headers: headers, context: context) {
-    if (method.isEmpty) throw ArgumentError('method cannot be empty.');
+    if (method.isEmpty) throw UriArgumentError('method cannot be empty.');
```
kevmoo: Should we leave this as `ArgumentError`? This is not an issue w/ the URI.

CHANGELOG.md
kevmoo: Better idea. Remove this new class. Throw `ArgumentError` BUT SET THE `name` param to `requestedUri`.

Then do `on ArgumentError catch (e, stackTrace)` and check the `name` for `requestedUri`.

Right?

isoos: I like it! Updated the code, but I found it simpler to use a generic catch and do the type check afterwards (1 code block instead of 3 very similar one).

kevmoo: SGTM

lib/shelf_io.dart
kevmoo: SGTM

lib/shelf_io.dart
```diff
@@ -74,8 +74,14 @@ Future handleRequest(HttpRequest request, Handler handler) async {
...
-    var response =
-        _logTopLevelError('Error parsing request.\n$error', stackTrace);
+    final isClientError = error is ArgumentError &&
+        (error.name == 'method' || error.name == 'requestedUri');
+    _logTopLevelError('Error parsing request.\n$error', stackTrace);
+    final response = isClientError
+        ? Response(400,
+            body: 'Bad Request',
+            headers: {HttpHeaders.contentTypeHeader: 'text/plain'})
+        : Response.internalServerError();
```
jonasfj: Maybe we should unwrap this a bit...
```dart
try {
} on ArgumentError catch (error) {
  if (error.name == 'method' || error.name == 'requestedUri') {
    .... 400
   return;
  }
  .... 500;
  return;
} catch (error, stackTrace) {
  var response =
        _logTopLevelError('Error parsing request.\n$error', stackTrace);
  await _writeResponse(response, request.response);
} 
```
But catching an `Error` subclass is almost always wrong. The correct thing to do would be to duplicate the checks... As we can't throw a different `Exception`, since `ArgumentError` is the correct thing to throw if invariants are violated in a public constructor. This is nits... 

But doing `} on ArgumentError catch (error) {` would clean up this a lot...

lib/shelf_io.dart
```diff
@@ -74,8 +74,14 @@ Future handleRequest(HttpRequest request, Handler handler) async {
   try {
     shelfRequest = _fromHttpRequest(request);
   } catch (error, stackTrace) {
-    var response =
-        _logTopLevelError('Error parsing request.\n$error', stackTrace);
+    final isClientError = error is ArgumentError &&
+        (error.name == 'method' || error.name == 'requestedUri');
+    _logTopLevelError('Error parsing request.\n$error', stackTrace);
```
jonasfj: We shouldn't log a top-level error if this was a client error.

isoos: If the package would use `package:logging` we could use a different log level, but with this one, I think it is better to log top level then omit it.

lib/shelf_io.dart
isoos: Updated the catch block. I'm not convinced that it adds enough value to warrant it, but I agree, it is a more correct version of the same code.

kevmoo:
@isoos – do you have the power to squash&merge?

isoos:
> @isoos – do you have the power to squash&merge?

eh, no :(

PR #125, opened 2019-12-05
Small doc and style cleanup

natebosch:
Fix some docs to be noun phrases for classes.

Remove some docs that are wholly redundant against the signature.

Remove an unnecessary private typedef.

Rename some methods that start with "get".

### DIFF
--- a/lib/src/message.dart
+++ b/lib/src/message.dart
@@ -12,7 +12,7 @@ import 'body.dart';
 import 'shelf_unmodifiable_map.dart';
 import 'util.dart';
 
-Body getBody(Message message) => message._body;
+Body extractBody(Message message) => message._body;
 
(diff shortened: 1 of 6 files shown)


lib/src/message.dart
```diff
@@ -153,7 +153,7 @@ Map<String, String> _adjustHeaders(Map<String, String> headers, Body body) {
   var sameEncoding = _sameEncoding(headers, body);
   if (sameEncoding) {
     if (body.contentLength == null ||
-        findHeader(headers, 'content-length') == body.contentLength.toString()) {
+        findHeader(headers, 'content-length') == '${body.contentLength}') {
```
kevmoo: 🤷‍♂

natebosch: kept it as 1 line instead of splitting

PR #129, opened 2020-02-25
Document that Request.ifModifiedSince throws FormatException

jonasfj:
It's a bit surprising that `Request.ifModifiedSince` throws `FormatException` if the incoming request is invalid.

We could also return `null`. But this has the downside of building a server that accepts invalid requests. The problem with accepting invalid requests is that you're going to break your users if you fix such a bug later.

Ideally, we should throw an exception that is caught and converted to `400` at some higher level.
But this would be a new way of structuring apps in `shelf`, so we should probably consider this in detail before we jump on the idea.

### DIFF
--- a/lib/src/request.dart
+++ b/lib/src/request.dart
@@ -68,6 +68,9 @@ class Request extends Message {
   ///
   /// This is parsed from the If-Modified-Since header in [headers]. If
   /// [headers] doesn't have an If-Modified-Since header, this will be `null`.
+  ///
+  /// Throws [FormatException], if incoming HTTP request has an invalid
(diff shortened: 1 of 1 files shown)


kevmoo:
Cleary an oversight. To throw "at some higher level" you'd almost want to parse these, if found, during the request creation. Honestly, you could do that now – it could be a feature of the `io` adapter to validate incoming requests for such things...

jonasfj:
> Honestly, you could do that now – it could be a feature of the io adapter to validate incoming requests for such things...

That might be nice to do.

PR #130, opened 2020-03-04
Handle cases where colon is in the first path segment

jonasfj:
`Request.url.path` is supposed to be relative to `Request.handlerPath`, this invariant is enforced by throwing `ArgumentError` in the constructor. However, if an incoming request has colon in the first
path segment the `shelf_io` adapter will create a request with `/` handler path creating a `Request.url.path` that is relative to `/` and which has colon in the first segment. This causes unnecessary encoding of the colon which breaks the invariant check.

In hindsight it's probably not ideal to use `Uri` for holding a relative path as this will always mangle the colon. But fixing this is very breaking.

I don't know if this check is likely to break other cases, I added a few test cases to ensure that slash encoding was also covered.

----
note. I don't really like this solution. But after playing with it I'm starting to be convinced that `Request.url` should instead be split into `Request.path` and `Request.query`, such that we don't need to have  `Uri` that is relative. But instead have a relative path exposed a `String` in `Request.path` or something like that. A change like this would, however, break everybody using shelf and reading the path or querystring params.

I guess we could annotate `Request.url` deprecated and gradually introduce `Request.path` and `Request.query`, if we're sure this is how we want to expose it. A relative path string might work a little better, but it'll be hard to access decoded path segments correctly (one would have to split on slashes and component decode each segment).

Not to mention that there could be other complications I haven't even considered :)

### DIFF
--- a/lib/src/request.dart
+++ b/lib/src/request.dart
@@ -4,6 +4,7 @@
 
 import 'dart:convert';
 
+import 'package:collection/collection.dart';
 import 'package:http_parser/http_parser.dart';
 import 'package:stream_channel/stream_channel.dart';
 
@@ -182,9 +183,21 @@ class Request extends Message {
     // cannot actually combine this.handlerPath and this.url.path, but we can
     // compare the pathSegments. In practice exposing this.url.path as a Uri
     // and not a String is probably the underlying flaw here.
-    final pathSegments = Uri(path: this.handlerPath).pathSegments.join('/') +
-        this.url.pathSegments.join('/');
-    if (pathSegments != this.requestedUri.pathSegments.join('/')) {
+    final handlerPathSegments = Uri(path: this.handlerPath).pathSegments;
+    // If there is a last pathSegment and it is empty we remove it because it
+    // is only present to retain a trailing slash.
+    Iterable<String> pathSegments = handlerPathSegments;
+    if (handlerPathSegments.isNotEmpty && handlerPathSegments.last.isEmpty) {
+      pathSegments = handlerPathSegments.getRange(
+        0,
+        handlerPathSegments.length - 1,
+      );
+    }
+    pathSegments = pathSegments.followedBy(this.url.pathSegments);
+    if (!IterableEquality().equals(
+      pathSegments,
+      this.requestedUri.pathSegments,
+    )) {
       throw ArgumentError.value(
           requestedUri,
           'requestedUri',
--- a/lib/src/request.dart
+++ b/lib/src/request.dart
@@ -178,11 +178,17 @@ class Request extends Message {
           requestedUri, 'requestedUri', 'may not have a fragment.');
     }
 
-    if (this.handlerPath + this.url.path != this.requestedUri.path) {
+    // Notice that because relative paths must encode colon (':') as %3A we
+    // cannot actually combine this.handlerPath and this.url.path, but we can
+    // compare the pathSegments. In practice exposing this.url.path as a Uri
+    // and not a String is probably the underlying flaw here.
+    final pathSegments = Uri(path: this.handlerPath).pathSegments.join('/') +
+        this.url.pathSegments.join('/');
+    if (pathSegments != this.requestedUri.pathSegments.join('/')) {
       throw ArgumentError.value(
           requestedUri,
           'requestedUri',
-          'handlerPath "$handlerPath" and url "$url" must '
+          'handlerPath "${this.handlerPath}" and url "${this.url}" must '
               'combine to equal requestedUri path "${requestedUri.path}".');
     }
   }
--- a/test/request_test.dart
+++ b/test/request_test.dart
@@ -33,6 +33,22 @@ void main() {
         expect(request.url, equals(Uri.parse('foo/bar?q=1')));
       });
 
+      test('may contain colon', () {
+        var request = Request('GET', Uri.parse('http://localhost/foo/bar:42'));
+        expect(request.url, equals(Uri.parse('foo/bar:42')));
+      });
+
+      test('may contain colon in first segment', () {
+        var request = Request('GET', Uri.parse('http://localhost/foo:bar/42'));
+        expect(request.url, equals(Uri.parse('foo%3Abar/42')));
+      });
+
+      test('may contain slash', () {
+        var request =
+            Request('GET', Uri.parse('http://localhost/foo/bar%2f42'));
+        expect(request.url, equals(Uri.parse('foo/bar%2f42')));
+      });
+
       test('is inferred from handlerPath if possible', () {
         var request = Request('GET', Uri.parse('http://localhost/foo/bar?q=1'),
             handlerPath: '/foo/');
--- a/test/shelf_io_test.dart
+++ b/test/shelf_io_test.dart
@@ -103,6 +103,14 @@ void main() {
     expect(response.body, 'Hello from /foo/bar');
   });
 
+  test('Request can handle colon in first path segment', () async {
+    await _scheduleServer(syncHandler);
+
+    var response = await _get(path: 'user:42');
+    expect(response.statusCode, HttpStatus.ok);
+    expect(response.body, 'Hello from /user:42');
+  });
+
   test('chunked requests are un-chunked', () async {
     await _scheduleServer(expectAsync1((request) {
       expect(request.contentLength, isNull);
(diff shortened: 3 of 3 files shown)


natebosch:
I'm having a hard time understanding when this would come up. Can you give an example? Did you hit this in a real use case?

> A change like this would, however, break everybody using shelf and reading the path or querystring params.

This package will see a breaking version bump for null safety - we can consider other changes like this as long as the migration isn't too hard - bonus points if it's a static failure to not migrate which it sounds like this would be.

jonasfj:
> I'm having a hard time understanding when this would come up. Can you give an example?

`GET 'http://localhost:8080/user:42'` will cause the request handler to throw an exception.
I added test case to `shelf_io_test.dart` because this is where you can trigger it.

While `GET 'http://localhost:8080/users/id:42'` works fine (so long as you don't mount a handler at `/users/` because that would trigger relative `Uri` that contains colon in the first path segment.

If we encoded the colon as `%3A` then `GET 'http://localhost:8080/user%3A42'` works fine.

> Did you hit this in a real use case?

I only discovered this because we had some log entries that requested: `https://pub.dev/https:%2F%2Fpub.dev%2F` which triggered this bug. I'm sure this is because we have a broken link somewhere. But a good first step is producing a 404 instead of a 500, in particular when the exception is thrown in shelf.

Excpetions thrown by shelf adapters are easily missed because request logging is often implemented at a higher level :)

> we can consider other changes like this as long as the migration isn't too hard - bonus points if it's a static failure to not migrate which it sounds like this would be.

That might be interesting to do. I'm not sure I've consider all aspects of changing the API. It just seems that colon encoding makes `Uri` a bad choice for representation of relative paths.

jonasfj:
@natebosch, no rush, but it'd be nice to get this bug fixed :)

lib/src/request.dart
```diff
@@ -175,11 +175,17 @@ class Request extends Message {
           requestedUri, 'requestedUri', 'may not have a fragment.');
     }
 
-    if (this.handlerPath + this.url.path != this.requestedUri.path) {
+    // Notice that because relative paths must encode colon (':') as %3A we
+    // cannot actually combine this.handlerPath and this.url.path, but we can
+    // compare the pathSegments. In practice exposing this.url.path as a Uri
+    // and not a String is probably the underlying flaw here.
+    final pathSegments = Uri(path: this.handlerPath).pathSegments.join('/') +
```
natebosch: Should this be `Uri(path: handlerPath).pathSegments.followedBy(url.pathSegments).join('/')`? Do we intend for the last segment of `handlerPath` to get merged with the first segment in `url` without a slash between them?

lib/src/request.dart
```diff
@@ -175,11 +175,17 @@ class Request extends Message {
...
+        this.url.pathSegments.join('/');
+    if (pathSegments != this.requestedUri.pathSegments.join('/')) {
       throw ArgumentError.value(
           requestedUri,
           'requestedUri',
-          'handlerPath "$handlerPath" and url "$url" must '
+          'handlerPath "${this.handlerPath}" and url "${this.url}" must '
```
natebosch: Why are we adding `this.`? I would expect it to trigger a lint...

jakemac53: looks like this is part of the constructor body so its actually disambiguating, presumably this was a bug before 😭

natebosch: 😮

lib/src/request.dart
natebosch: @jonasfj - is this still something you want to look at?

lib/src/request.dart
jonasfj: Yes, minor bug that produced hard to debug error messages.

lib/src/request.dart
jonasfj: Sorry, I'm slow... thanks for the ping :)

```dart
    final pathSegments = Uri(path: this.handlerPath)
        .pathSegments
        .followedBy(this.url.pathSegments)
        .join('/');
```
Won't work because `handlerPath` always ends in `/`.

Hence, if `handlerPath` is `/static/` and `url` is `dir/file.html` we would get `static//dir/file.html` instead of `/static/dir/file.html` (which is the request path). I suppose `static/` gives rise to an empty path segment to preserve trailing slash.

-----

If I understand it right `handlerPath` is always a prefix of `requestedUri.path`. What we are trying to do is check that `this.url.path` is the remainder of the path. This impossible because making a suffix of `requestedUri.path` into a relative path encoded with `Uri` causes encoding of colon (which is destructive, we can't tell the difference between colon and `%3A`).

But since this is just an assertion, not something that should happen at all. I think we can ignore the case where `requestedUri` contains an encoded character and `uri` contains the decoded counter part. It's technically possible to create this invariant violation with the `Request` constructor.

So I essentially opted to relax the invariant check by always decoding before we compare. I think this is better than having valid requests cause internal errors.

jonasfj: We could also drop this invariant check completely :see_no_evil: 

But I think it can catch some serious mistakes, and that it can still do this. Even if we relaxed it a bit.

natebosch: I think we want to be using `package:path` or maybe `IterableEquality`. We already depend on both `path` and `collection` packages. I'm very nervous about this manual string concatenation.

We can get the path segments without the last empty one, and then append `this.url.pathSegments`.

```dart
    final pathSegments = Uri(path: handlerPath)
        .pathSegments
        .takeWhile((s) => s.isNotEmpty)
        .followedBy(this.url.pathSegments);
```

Then compare with either

```dart
    if (p.url.joinAll(pathSegments) != p.url.joinAll(this.requestedUri.pathSegments)) {
```

or

```dart
    if (const IterableEquality().equals(pathSegments, this.requestedUri.pathSegments)) {
```

natebosch:
> I'm starting to be convinced that `Request.url` should instead be split into `Request.path` and `Request.query`, such that we don't need to have `Uri` that is relative. But instead have a relative path exposed a `String` in `Request.path` or something like that. A change like this would, however, break everybody using shelf and reading the path or querystring params.
> 
> I guess we could annotate `Request.url` deprecated and gradually introduce `Request.path` and `Request.query`, if we're sure this is how we want to expose it. A relative path string might work a little better, but it'll be hard to access decoded path segments correctly (one would have to split on slashes and component decode each segment).

What do you think about introducing `Request.pathSegments` instead of `Request.path` so there is no concern about splitting and decoding?

lib/src/request.dart
jonasfj: >         .takeWhile((s) => s.isNotEmpty)
Are you sure a path with multiple slashes `/folder///subfolder/file.txt` won't have empty segments.

jonasfj:
> What do you think about introducing `Request.pathSegments` instead of `Request.path` so there is no concern about splitting and decoding?

I think users will want to have a `.path` somewhere that returns a string, I imagine most novice users don't care about whether it's `%f2` or `/` :)
But if we made `Request.pathSegments` canonical, and just had `Request.path` as a getter that would convert to `String` on-the-fly I don't think `Request.pathSegments` is much better.

lib/src/request.dart
natebosch: I do think it would be unusual to have a `handlerPath` with empty segments...

pubspec.yaml
```diff
@@ -5,7 +5,7 @@ description: >-
 homepage: https://github.com/dart-lang/shelf
 
 environment:
-  sdk: ">=2.1.0 <3.0.0"
+  sdk: ">=2.6.0 <3.0.0"
```
natebosch, line 8: Why did this need to change?

It will need to be changed in `.travis.yml` too if we need to bump the min SDK.

natebosch:
> > What do you think about introducing `Request.pathSegments` instead of `Request.path` so there is no concern about splitting and decoding?
> 
> I think users will want to have a `.path` somewhere that returns a string, I imagine most novice users don't care about whether it's `%f2` or `/` :)
> But if we made `Request.pathSegments` canonical, and just had `Request.path` as a getter that would convert to `String` on-the-fly I don't think `Request.pathSegments` is much better.

It would avoid the temptation to treat `path` as a `Uri`. The safest thing to do, I think, is to parse into segments once at the root, and then have all other methods operate on those already parsed segments.

pubspec.yaml
jonasfj: I completely forgot why, maybe I removed the bits that required `2.6.0`.

PR #132, opened 2020-04-10
Fix failure test with breaking HTTP Header change

ZichangG:
This is a follow-up cl(https://dart-review.googlesource.com/c/sdk/+/143081) for previous HttpHeaders casing issue.

With this change, HttpParser will preserve the casing of HttpHeader fields name.

To avoid the breakage, I comment out the line once the cl lands. It can be added back.

### DIFF
--- a/test/shelf_io_test.dart
+++ b/test/shelf_io_test.dart
@@ -221,7 +221,9 @@ void main() {
 
     var response = await _post(body: 'Hello');
     expect(response.statusCode, HttpStatus.notFound);
-    expect(response.headers['date'], 'Mon, 23 May 2005 22:38:34 GMT');
+    // TODO: after https://dart-review.googlesource.com/c/sdk/+/143081 lands,
+    // uncomment the following line.
+    // expect(response.headers['Date'], 'Mon, 23 May 2005 22:38:34 GMT');
     expect(
         response.stream.bytesToString(), completion(equals('Hello, world!')));
   });
--- a/test/shelf_io_test.dart
+++ b/test/shelf_io_test.dart
@@ -209,7 +209,7 @@ void main() {
         expect(channel.stream.first, completion(equals('Hello'.codeUnits)));
 
         channel.sink.add(('HTTP/1.1 404 Not Found\r\n'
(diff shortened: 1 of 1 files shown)


test/shelf_io_test.dart
```diff
@@ -221,7 +221,9 @@ void main() {
 
     var response = await _post(body: 'Hello');
     expect(response.statusCode, HttpStatus.notFound);
-    expect(response.headers['date'], 'Mon, 23 May 2005 22:38:34 GMT');
+    // TODO: after https://dart-review.googlesource.com/c/sdk/+/143081 lands,
```
kevmoo, line 224: Need to be careful here. We test this against multiple SDKs.

natebosch: Does it work to make it lowercase on line 212 instead?

ZichangG: > Need to be careful here. We test this against multiple SDKs.

OK. Any additional tests I need to run?
> Does it work to make it lowercase on line 212 instead?

No. Http parser will not enforce lowercased field and the key will be `Date`(The same as the input).

natebosch: It will match the case of what was given to it right? Right now we are giving a capital `Date` - if we give it a lowercase `date` would it keep it?

ZichangG: Exactly. It will keep!

natebosch: Line 212 is where it is effectively passed in. If we make that lowercase and leave this line as is I think we'll be backwards and forwards compatible.

ZichangG: Good point! That's even better. Are you able to modify the pr here or I have to create another pr?

natebosch: I updated the PR, you should be able to `git pull` to see it locally.

Can you run against your patched SDK to confirm, or do you want me to pull down the change locally?

ZichangG: It worked with my change.

natebosch: Awesome. Thanks!

Issue #29, opened 2015-06-05
pkg/shelf: Implement Cookie() utilities


DartBot:
_Originally opened as dart-lang/sdk#18845_

*This issue was originally filed by warren.st...&#064;gmail.com*
_____
It would be really useful to have utilities for managing cookies (parsing cookies, setting cookies)
&nbsp;

DartBot:
<a href="https://github.com/kevmoo"><img src="https://avatars.githubusercontent.com/u/17034?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [kevmoo](https://github.com/kevmoo)**

----

*Removed [Type-Defect](../labels/Type-Defect) label.*
*Added [Type-Enhancement](../labels/Type-Enhancement), [Area-Pkg](../labels/Area-Pkg), [Pkg-Shelf](../labels/Pkg-Shelf), [Triaged](../labels/Triaged) labels.*

DartBot:
<a href="https://github.com/kevmoo"><img src="https://avatars.githubusercontent.com/u/17034?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [kevmoo](https://github.com/kevmoo)**

----

See also https://code.google.com/p/dart/issues/detail?id=19005

DartBot:
<a href="https://github.com/Andersmholmgren"><img src="https://avatars.githubusercontent.com/u/698895?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [Andersmholmgren](https://github.com/Andersmholmgren)**

----

I was going to take a crack at adding cookie based session support to shelf_auth but then ran into this issue with lack of cookie support in shelf so I had to shelve the idea (no pun intended. OK maybe a little)

DartBot:
*This comment was originally written by warren.strange...&#064;gmail.com*
_____
Anders - I implemented very basic session support by copying the dart:io cookie code. See https://github.com/wstrange/shelf_simple_session

DartBot:
<a href="https://github.com/Andersmholmgren"><img src="https://avatars.githubusercontent.com/u/698895?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [Andersmholmgren](https://github.com/Andersmholmgren)**

----

[...]

PR #134, opened 2020-05-10
Add cookie support

allentv:
This is an intial effort to add support for cookies. I would like to get some feedback about this approach after which I can add tests.

### DIFF
--- a/lib/src/request.dart
+++ b/lib/src/request.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:convert';
+import 'dart:io';
 
 import 'package:http_parser/http_parser.dart';
 import 'package:stream_channel/stream_channel.dart';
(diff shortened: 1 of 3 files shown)


allentv:
@googlebot I signed it!


Regards,
Allen Thomas Varghese
+353 89 965 7732

LinkedIn : http://ie.linkedin.com/in/allentv
Twitter : https://twitter.com/allentv4u

allentv:
This PR should implement some part of https://github.com/dart-lang/shelf/issues/29

lib/src/request.dart
```diff
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:convert';
+import 'dart:io';
```
kevmoo, line 6: Sadly, we can't land this. We don't want `dart:io` "things" leaking into the core library. It's meant to work without `dart:io`

kevmoo:
This approach won't work. We don't want `dart:io` types leaking into anything exposed by `shelf.dart`

lib/src/request.dart
allentv: Thanks for the review. Appreciate it!

PR #135, opened 2020-05-13
Add cookie support - Take 2

allentv:
This is my second attempt at adding Cookie support. First attempt was using dart.io. This time I skipped the io package and added separate implementation. Let me know if this solution looks good. Based on your feedback, I can add more code and tests.


allentv:
Closing this in favor of https://github.com/izolate/shelf-cookie

PR #138, opened 2020-06-05
Supporting multiple header values in Request and Response.

### LINKED ISSUE #44: Add support for multi-valued headers.

isoos:
- Fixes #44.
- Manually tested that `set-cookie` header is working (`package:http` does not support multiple values).
- The `headers` map is created lazily by calling `.join(',')` on the values. (Consistent in both request and response object, preserves current behaviour when multiple header values are received.)

### DIFF
--- a/lib/src/headers.dart
+++ b/lib/src/headers.dart
@@ -1,0 +1,57 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:collection';
+
+import 'package:http_parser/http_parser.dart';
+import 'package:shelf/src/util.dart';
+
+final _emptyHeaders = Headers._empty();
+
+/// Unmodifiable, key-insensitive header map.
+class Headers extends UnmodifiableMapView<String, List<String>> {
+  Map<String, String> _singeValues;
+
+  factory Headers.from(Map<String, List<String>> values) {
+    if (values == null || values.isEmpty) {
+      return _emptyHeaders;
+    } else if (values is Headers) {
+      return values;
+    } else {
+      return Headers._(values);
+    }
+  }
+
+  Headers._(Map<String, List<String>> values)
+      : super(
+          CaseInsensitiveMap<List<String>>.from(
+            Map<String, List<String>>.fromEntries(
+              values.entries
+                  .where((e) => e.value != null && e.value.isNotEmpty)
+                  .map(
+                    (e) => MapEntry<String, List<String>>(
+                      e.key,
+                      List.unmodifiable(e.value),
+                    ),
+                  ),
+            ),
+          ),
+        );
+
+  Headers._empty() : super(<String, List<String>>{});
+  factory Headers.empty() => _emptyHeaders;
+
+  Map<String, String> get singleValues {
+    return _singeValues ??= _SingleValueHeaders(
+      CaseInsensitiveMap<String>.from(
+        map((key, value) =>
+            MapEntry<String, String>(key, joinHeaderValues(value))),
+      ),
+    );
+  }
+}
+
+class _SingleValueHeaders extends UnmodifiableMapView<String, String> {
+  _SingleValueHeaders(Map<String, String> map) : super(map);
+}
--- a/lib/src/message.dart
+++ b/lib/src/message.dart
@@ -26,6 +26,7 @@ abstract class Message {
   final Headers _headers;
 
   /// The HTTP headers.
+  /// The keys in this Map are normalized, and access is case-insensitive.
   ///
   /// If a header occurs more than once in the query string, they are mapped to
   /// by concatenating them with a comma.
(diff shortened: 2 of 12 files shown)


lib/src/request.dart
```diff
@@ -126,7 +126,7 @@ class Request extends Message {
   // TODO(kevmoo) finish documenting the rest of the arguments.
   Request(String method, Uri requestedUri,
       {String protocolVersion,
-      Map<String, String> headers,
+      Map<String, /* String | List<String> */ dynamic> headers,
```
kevmoo, line 129: Wonder if we should use `Object` here and below, since we don't want to allow `null`, right? (Thinking forward to NNBD)

isoos: As an empty `<String>[]` can be used to remove an existing header with the `.change()` call, I don't mind not using `null`.

isoos: Updated, now the code does not accept `null`.

kevmoo:
@isoos – you'll have to rebase

@jonasfj @natebosch – any thoughts here?

lib/src/message.dart
```diff
@@ -9,22 +9,38 @@ import 'package:collection/collection.dart';
...
+  /// The returned map is unmodifiable.
+  Map<String, String> get headers => _headers.singleValues;
+
+  /// The HTTP headers with multiple values.
+  ///
+  /// If a header occurs only once, its value is a singleton list.
+  /// If a header occurs with no value, the empty string is used as the value
+  /// for that occurrence.
+  ///
+  /// The returned map and the lists it contains are unmodifiable.
```
jonasfj, line 42: Unrelated to this PR, but it would be really lovely if the documentation said that headers were case-insensitive :)

test/shelf_io_test.dart
```diff
@@ -137,6 +137,22 @@ void main() {
...
+  test('multiple headers are received from the client', () async {
+    await _scheduleServer((request) {
+      return Response.ok('Hello from /', headers: {
+        'requested-values': request.headersAll['request-values'],
+        'requested-values-length':
+            request.headersAll['request-values'].length.toString(),
+      });
+    });
+
+    // TODO: update test once `package:http` is able to send multiple header values
```
jonasfj, line 149: Can't we do it with `dart:io`?

https://api.dart.dev/stable/2.8.4/dart-io/HttpClient/get.html

Not sure if it concatenates the headers. Otherwise, we could do it with a TCP socket, if it's just a `GET` request it might not be that horrible a hack :see_no_evil:

lib/src/headers.dart
```diff
@@ -0,0 +1,57 @@
...
+                  .where((e) => e.value != null && e.value.isNotEmpty)
+                  .map(
+                    (e) => MapEntry<String, List<String>>(
+                      e.key,
+                      List.unmodifiable(e.value),
+                    ),
+                  ),
+            ),
+          ),
+        );
```
natebosch, line 40:
```suggestion
      : super(CaseInsensitiveMap<List<String>>.from({
          for (var e in values.entries)
            if (e.value != null && e.value.isNotEmpty)
              e.key: List.unmodifiable(e.value)
        }));
```

It is fine to bump the min SDK to allow this.

lib/src/headers.dart
```diff
@@ -0,0 +1,57 @@
...
+
+  Headers._empty() : super(<String, List<String>>{});
+  factory Headers.empty() => _emptyHeaders;
+
+  Map<String, String> get singleValues {
+    return _singeValues ??= _SingleValueHeaders(
+      CaseInsensitiveMap<String>.from(
+        map((key, value) =>
+            MapEntry<String, String>(key, joinHeaderValues(value))),
```
natebosch, line 49: I would expect that `<String, String>` can be filled in by inference here.

```suggestion
            MapEntry(key, joinHeaderValues(value))),
```

lib/src/util.dart
```diff
@@ -72,3 +76,31 @@ bool hasHeader(Map<String, String> headers, String name) {
...
+
+List<String> expandHeaderValue(dynamic v) {
+  if (v is String) {
+    return [v];
+  } else if (v is List<String>) {
+    return v;
+  } else {
+    throw ArgumentError('Expected String or List<String>, got: `$v`.');
+  }
+}
```
jonasfj, line 98: The issue @natebosch mentions looks to origin from here... 
```
Error thrown by handler.
Invalid argument(s): Expected String or List<String>, got: `null`.

package:shelf/src/response.dart 66:9                             new Response.ok
```
In the current release the following code works:
```dart
  test('allows null header', () {
    var response = Response.ok('hello, world', headers: {'x-hack': null});
    expect(response.readAsString(), completion(equals('hello, world')));
  });
```
I don't think that'll work on this branch as we'll throw an `ArgumentError` here.

lib/src/message.dart
isoos: Done.

test/shelf_io_test.dart
isoos: Done.

lib/src/request.dart
isoos: Reverted back to accept `null`, but kept `Object`.

isoos:
PTAL.

PR #143, opened 2020-06-19
Fix #142 by comparing pathSegments as / joined strings

### LINKED ISSUE #142: `Request.change` broken with latest release
See https://travis-ci.org/github/dart-lang/build/jobs/699942234.

This looks like it was caused by https://github.com/dart-lang/shelf/pull/130.

This is the specific code that is now failing https://github.com/dart-lang/build/blob/554d93df198c53adff9d8d85c043c0ad8ddcdcfe/build_runner/lib/src/server/server.dart#L115.

This is taking an incoming request whose handler path is `"/"` and whose `uri.path` is `"$graph/"`, and then calling `request.change(path: "$graph")` on that.

According to the docs all that is required is that `path` is a prefix of `uri.path` https://pub.dev/documentation/shelf/latest/shelf/Request/change.html.

jonasfj:
As proposed in https://github.com/dart-lang/shelf/pull/130/commits/b7da1fd85a159aaea5943d9cadac3efccc8535d7 which we changed to use `IterableEquality().equals` before we landed #130 -- we probably failed to consider all the corner cases.

We could probably also use `IterableEquality().equals` if we add a few more special cases, I haven't exactly worked out what would be needed to make that work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.7-dev
+
+* Fix internal error in `Request` when `.change()` matches the full path.
+
 ## 0.7.6
 
 *   Supports multiple header values in `Request` and `Response`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.7-dev
+## 0.7.7
 
 * Fix internal error in `Request` when `.change()` matches the full path.
 
--- a/lib/src/request.dart
+++ b/lib/src/request.dart
@@ -183,21 +183,9 @@ class Request extends Message {
     // cannot actually combine this.handlerPath and this.url.path, but we can
     // compare the pathSegments. In practice exposing this.url.path as a Uri
     // and not a String is probably the underlying flaw here.
-    final handlerPathSegments = Uri(path: this.handlerPath).pathSegments;
-    // If there is a last pathSegment and it is empty we remove it because it
-    // is only present to retain a trailing slash.
-    Iterable<String> pathSegments = handlerPathSegments;
-    if (handlerPathSegments.isNotEmpty && handlerPathSegments.last.isEmpty) {
-      pathSegments = handlerPathSegments.getRange(
-        0,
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.7.7-dev
```
kevmoo, line 1: Should we just drop this as `0.7.7` NOW so we can publish?

natebosch:
I don't think this fixes the failure in `build_runner`.

jonasfj:
Seems to fix it on my end.. at-least for the one test case :)

```
[build_runner]$pub run test test/server/serve_integration_test.dart
00:07 +6 -1: /$graph /$graph/ should (try to) send the HTML page [E]                                                                                                         
  Expected: <Instance of 'AssetNotFoundException'>
    Actual: ArgumentError:<Invalid argument (requestedUri): handlerPath "/$graph/" and url "" must combine to equal requestedUri path "/$graph/".: Instance of '_SimpleUri'>
     Which: is not an instance of 'AssetNotFoundException'
  
  package:test_api                                expect
  test/server/serve_integration_test.dart 136:11  main.<fn>.<fn>
  
00:07 +17 -1: Some tests failed.                                                                                                                                             
[build_runner]$pub get
Resolving dependencies... 
Warning: You are using these overridden dependencies:
! shelf 0.7.7 from path /source/shelf
Changed 1 dependency!
[build_runner]$pub run test test/server/serve_integration_test.dart
00:05 +18: All tests passed!                                                                                                                                                 
[build_runner]$
```

jonasfj:
For the record we could also have done:
```dart
-    pathSegments = pathSegments.followedBy(this.url.pathSegments);
+    pathSegments = pathSegments.followedBy(
+      this.url.pathSegments.isEmpty &&
+              handlerPathSegments.isNotEmpty &&
+              handlerPathSegments.last.isEmpty
+          ? ['']
+          : this.url.pathSegments,
+    );
```
But adding another special rule to the `pathSegments` comparison didn't seem wise :)

PR #146, opened 2020-08-12
Fix handlerPath check when a path contains //

natebosch:
If the requested URI is not canonical and contains a `//` separator the
previous check would not allow changing a request to set the
`handlerPath` to the part of the path leading up to the `//`. This shows
up in some requests used in tests for the VM service.

### DIFF
--- a/test/request_test.dart
+++ b/test/request_test.dart
@@ -365,6 +365,16 @@ void main() {
         expect(copy.url, Uri.parse(''));
       });
 
+      test('allows changing pth to URL leading to double //', () {
+        var uri = Uri.parse('https://test.example.com/some_base//more');
+        var request = Request('GET', uri,
+            handlerPath: '', url: Uri.parse('some_base//more'));
+
+        var copy = request.change(path: 'some_base');
+        expect(copy.handlerPath, '/some_base/');
+        expect(copy.url, Uri.parse('/more'));
+      });
+
       test('throws if path does not match existing uri', () {
         var uri = Uri.parse('https://test.example.com/static/dir/file.html');
         var request = Request('GET', uri,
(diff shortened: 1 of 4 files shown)


natebosch:
cc @jonasfj - this is another edge case that surfaced in the SDK and was broken by the changes you made to allow `:` characters. This further reinforces that the string/URI manipulation is risky.

test/request_test.dart
```diff
@@ -365,6 +365,16 @@ void main() {
         expect(copy.url, Uri.parse(''));
       });
 
+      test('allows changing pth to URL leading to double //', () {
```
kevmoo, line 368: `path`

jonasfj:
Nice catch!

Yeah, I agree.. fiddling with URLs is really hard.
Maybe we're trying to be a little too high-level... It's a hard balance to walk.

natebosch:
Here is an example failure I'm hoping this fixes:

https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket.appspot.com/8872685480086439280/+/steps/test_results/0/logs/new_test_failures__logs_/0

```
Invalid argument (requestedUri): handlerPath "/HJbVrcIemLk=/" and url "/ws" must combine to equal requestedUri path "/HJbVrcIemLk=//ws".: Instance of '_Uri'
package:shelf/src/request.dart 234:20                          Request.change
package:dds/src/dds_impl.dart 134:29                           _DartDevelopmentService._authCodeMiddleware.<fn>
package:shelf/shelf_io.dart 63:34                              serveRequests.<fn>.<fn>
```

jonasfj:
I think there is a solid chance it'll fix that, I recall the split+join on slashes that `Uri` does, play tricks with leading/trailing slashes.

PR #152, opened 2020-11-20
Migrate to null safety

kevmoo:


kevmoo:
I still need to do the version/changelog thing here.

Thoughts on header **values** being nullable?

It leaks in everywhere. I'm wondering if it's ever valid.

Thoughts?

kevmoo:
Updated the changelog

lib/src/message.dart
```diff
@@ -31,7 +31,7 @@ abstract class Message {
   /// by concatenating them with a comma.
   ///
   /// The returned map is unmodifiable.
-  Map<String, String> get headers => _headers.singleValues;
+  Map<String, String?> get headers => _headers.singleValues;
```
jonasfj, line 34: Would `Map<String, String?>` -> `Map<String, String>` break a lot of things in shelf?
(I guess I'd have to clone and play with to really see it).

side note: I kind of wish we deprecated `Message.headers` in favor of `Message.headersAll`, since the comma-separate-value is rather surprising.

kevmoo:
I don't see any path where null headers were ever really existing in
practice.

headersAll is annoying because most of the time, you just want one value

lib/src/message.dart
jonasfj: > I don't see any path where null headers were ever really existing in practice.

I agree, indeed `Headers.singleValues` has the type `Map<String, String>`, so the value here cannot (if you believe in null-safety) be `null`.

> `headersAll` is annoying because most of the time, you just want one value

I suppose that's true, and the failure scenario isn't super bad.

kevmoo: So...any other feedback?

lib/shelf_io.dart
```diff
@@ -96,26 +109,28 @@ Future handleRequest(HttpRequest request, Handler handler) async {
...
-        _logError(shelfRequest, 'Error thrown by handler.\n$error', stackTrace);
+    response = _logError(
+      shelfRequest,
+      'Error thrown by handler.\n$error',
+      stackTrace,
+    );
   }
 
-  if (response == null) {
-    await _writeResponse(_logError(shelfRequest, 'null response from handler.'),
```
jakemac53, line 115: Was this overly defensive (dead) code before? I generally don't expect so see this type of change in migrations to no longer handle null where it was previously handled.

lib/src/message.dart
jakemac53: I am not following why  this is not a `Map<String, String>`?

lib/shelf_io.dart
kevmoo: this was dead, I think. There are no code paths where `response` could be null. Finally caught it.

...at least I think!

lib/src/message.dart
kevmoo: Oh wait...it should be.

Weird...will look

kevmoo: just a mistake on my part!

kevmoo: fixed

kevmoo:
PTAL

jakemac53 (approved):

test/shelf_io_test.dart
```diff
@@ -43,7 +43,7 @@ void main() {
   });
 
   test('sync null response leads to a 500', () async {
-    await _scheduleServer((request) => null);
+    await _scheduleServer((request) => Response.internalServerError());
```
natebosch, line 46: This changed the intent of the test. It no longer matches the name.

Should it be removed? Should it be moved to an `opt_out_test.dart`?

kevmoo: Good note! Hrm....

lib/shelf_io.dart
natebosch: Running tests with `pub run --no-sound-null-safety` shows a failure. It looks related to this.

PR #154, opened 2020-12-02
Restore handling of null response

natebosch:
Opt-in code cannot pass a null response, however opt-out code can.
Retain check for a null `Response`, cast to `dynamic` to silence
analyzer warning.

Add a test that specifically opts out of null safety to test this
behavior.

### DIFF
--- a/test/shelf_io_test.dart
+++ b/test/shelf_io_test.dart
@@ -42,7 +42,6 @@ void main() {
     expect(response.body, 'Hello from /');
   });
 
-
   test('thrown error leads to a 500', () async {
     await _scheduleServer((request) {
       throw UnsupportedError('test');
--- a/test/shelf_io_test.dart
+++ b/test/shelf_io_test.dart
@@ -42,21 +42,6 @@ void main() {
     expect(response.body, 'Hello from /');
   });
 
-  test('sync null response leads to a 500', () async {
-    await _scheduleServer((request) => Response.internalServerError());
-
-    var response = await _get();
-    expect(response.statusCode, HttpStatus.internalServerError);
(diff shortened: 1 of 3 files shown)


natebosch:
This does copy/paste a fair bit of test utility code in that file. I wasn't sure if it was worth it to pull it out into the test utils - let me know if you think I should.

test/shelf_io_test.dart
```diff
@@ -42,21 +42,6 @@ void main() {
...
 
-  test('sync null response leads to a 500', () async {
-    await _scheduleServer((request) => Response.internalServerError());
-
-    var response = await _get();
-    expect(response.statusCode, HttpStatus.internalServerError);
-    expect(response.body, 'Internal Server Error');
-  });
-
-  test('async null response leads to a 500', () async {
```
kevmoo, line 53: isn't this a valid test in sound mode?

kevmoo (approved):

test/shelf_io_test.dart
natebosch: It passes for the wrong reason.

What this test demonstrates in sound mode is that the `Future.value` constructor has a bad API for null safety, and the `Future.value(null)` is acting like a `Future.error` in sound mode. We already have that behavior covered in sound mode by the test `'async error leads to a 500'`.

kevmoo: 👍

natebosch: I'll use `async => null` instead of `Future.value(null)` to make this more clear why it's only an opt-out test. The former would have a static error in opt-in code.

jakemac53 (approved):

PR #155, opened 2020-12-04
Add shim for old unsound null behavior

natebosch:
Keep the static type of the Headers map as non-null so that migrated
libraries need to make explicit the behavior for nulls rather than
relying on the hidden detail from `package:shelf`. Cast to `dynamic` to
avoid analysis warnings and check for nulls with a fallback that will
retain the behavior pre-migration of dropping the header.


grouma (approved):

grouma (approved):

natebosch:
@kevmoo are you OK with this approach or should we push to clean up cases as we find them?

kevmoo (approved):
I think it's fine.

PR #178, opened 2021-04-23
Avoid wrapping List<int>/Stream<List<int>> response bodies in a CastList/CastStream

jonahwilliams:
(Allows https://github.com/flutter/flutter/issues/81010 to be fixed, though we need to change some code on the flutter side too)

TLDR: when writing data to a socket in dart, anything besides a Uint8List or Int8List will be necessarily copied (c.f. https://github.com/dart-lang/sdk/blob/3f0ad61daaac68e591bf01d0a3096926c30c1d42/sdk/lib/io/common.dart#L92)

shelf will unnecessarily wrap all Lists in a CastList which will perform runtime checks for every single element.

The combination of these behaviors causes requests for large files in the flutter run web server to be incredibly slow, observed to be on the order of multiple minutes locally. With the combination of this change and a patch to the tool to avoid creating Uint8List views, the time to serve the app drops to a few seconds.

I took a CPU profile starting from when chrome opened until the app's first frame. I could upload the actual timeline data too, but a picture something something a thousand words

# Before:

Dart DevTools - Google Chrome 4_22_2021 5_50_04 PM

# After:

Dart DevTools - Google Chrome 4_22_2021 5_50_52 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.1.1
+
+* Avoid wraping response bodies that already contained `List<int>` in a `CastList`.
+  This will improve the performance of large response bodies
+
 ## 1.1.0
 
 * Change `Request.hijack` return type from `void` to `Never`. This may cause
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## 1.1.1
 
-* Avoid wraping response bodies that already contained `List<int>` in a `CastList`.
-  This will improve the performance of large response bodies
+* Avoid wraping response bodies that already contained `List<int>` or `Stream<List<int>>` in a `CastList`/`CastStream` to improve performance.
 
 ## 1.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 1.1.1
 
-* Avoid wraping response bodies that already contained `List<int>` or `Stream<List<int>>` in a `CastList`/`CastStream` to improve performance.
+* Avoid wrapping response bodies that already contained `List<int>` or
+  `Stream<List<int>>` in a `CastList`/`CastStream` to improve
+  performance.
 
 ## 1.1.0
 
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -5,8 +5,6 @@
 import 'dart:async';
 import 'dart:convert';
 
-import 'dart:typed_data';
-
 /// The body of a request or response.
 ///
 /// This tracks whether the body has been read. It's separate from [Message]
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -5,6 +5,8 @@
 import 'dart:async';
 import 'dart:convert';
 
+import 'dart:typed_data';
+
 /// The body of a request or response.
 ///
 /// This tracks whether the body has been read. It's separate from [Message]
@@ -52,6 +54,13 @@ class Body {
         contentLength = encoded.length;
         stream = Stream.fromIterable([encoded]);
       }
+    } else if (body is Uint8List) {
+      // Any data that is written to a Dart socket that is not a Uint8List or
+      // Int8List must be copied into a new list. Allow users to provide a
+      // typed data object to avoid this copy, which is important for performance
+      // of large file requests.
+      contentLength = body.length;
+      stream = Stream.fromIterable([body]);
     } else if (body is List) {
       contentLength = body.length;
       stream = Stream.fromIterable([body.cast()]);
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -53,15 +53,12 @@ class Body {
         stream = Stream.fromIterable([encoded]);
       }
     } else if (body is List<int>) {
-      // Any data that is written to a Dart socket that is not a Uint8List or
-      // Int8List must be copied into a new list. Allow users to provide a
-      // typed data object to avoid this copy, which is important for performance
-      // of large file requests.
+      // Avoid performance overhead from an unnecessary cast.
       contentLength = body.length;
-      stream = Stream.fromIterable([body]);
+      stream = Stream.value(body);
     } else if (body is List) {
       contentLength = body.length;
-      stream = Stream.fromIterable([body.cast()]);
+      stream = Stream.value(body.cast());
     } else if (body is Stream) {
       stream = body.cast();
     } else {
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -59,6 +59,9 @@ class Body {
     } else if (body is List) {
       contentLength = body.length;
       stream = Stream.value(body.cast());
+    } else if (body is Stream<List<int>>) {
+      // Avoid performance overhead from an unnecessary cast.
+      stream = body;
     } else if (body is Stream) {
       stream = body.cast();
     } else {
--- a/lib/src/body.dart
+++ b/lib/src/body.dart
@@ -54,7 +54,7 @@ class Body {
         contentLength = encoded.length;
         stream = Stream.fromIterable([encoded]);
       }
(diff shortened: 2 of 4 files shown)


lib/src/body.dart
```diff
@@ -52,6 +54,13 @@ class Body {
         contentLength = encoded.length;
         stream = Stream.fromIterable([encoded]);
       }
+    } else if (body is Uint8List) {
```
kevmoo, line 57: Or could this just be `is List<int>`?

jonahwilliams: Ahh yeah, changed!

lib/src/body.dart
```diff
@@ -52,6 +52,13 @@ class Body {
         contentLength = encoded.length;
         stream = Stream.fromIterable([encoded]);
       }
+    } else if (body is List<int>) {
+      // Any data that is written to a Dart socket that is not a Uint8List or
+      // Int8List must be copied into a new list. Allow users to provide a
+      // typed data object to avoid this copy, which is important for performance
+      // of large file requests.
+      contentLength = body.length;
+      stream = Stream.fromIterable([body]);
```
natebosch, line 61:
```suggestion
      stream = Stream.value(body);
```

grouma (approved):
Let's update the changelog and pubspec to prep for release.

jonahwilliams:
1.1.1 ?

lib/src/body.dart
```diff
@@ -52,6 +52,13 @@ class Body {
         contentLength = encoded.length;
         stream = Stream.fromIterable([encoded]);
       }
+    } else if (body is List<int>) {
+      // Any data that is written to a Dart socket that is not a Uint8List or
+      // Int8List must be copied into a new list. Allow users to provide a
+      // typed data object to avoid this copy, which is important for performance
+      // of large file requests.
```
natebosch, line 59: This comment feels too detailed to me - the extra detail isn't actionable to future maintainers. It's great context to have in the commit message.

I'd go with 

```dart
// Avoid performance overheads from an unnecessary cast.
```

test/response_test.dart
```diff
@@ -25,6 +26,30 @@ void main() {
     });
   });
 
+  test('supports a Uint8List body without copying', () async {
+    var bytes = Uint8List(10);
+    var response = Response.ok(bytes);
+
+    expect(response.contentLength, 10);
+    expect(identical(await response.read().first, bytes), true);
```
natebosch, line 34:
```suggestion
    expect(await response.read().single, same(bytes));
```

Using `single` guarantees there aren't extra elements. Using `same` over comparing against the literal `true` will improve error messages if this fails.

test/response_test.dart
```diff
@@ -25,6 +26,30 @@ void main() {
...
+  });
+
+  test('supports a List<int> body without copying', () async {
+    var bytes = <int>[1, 2, 3, 4];
+    var response = Response.ok(bytes);
+
+    expect(response.contentLength, 4);
+    expect(identical(await response.read().first, bytes), true);
```
natebosch, line 42:
```suggestion
    expect(await response.read().single, same(bytes));
```

test/response_test.dart
```diff
@@ -25,6 +26,30 @@ void main() {
...
+  });
+
+  test('Copies a dynamic list of int elements', () async {
+    var bytes = <dynamic>[1, 2, 3, 4];
+    var response = Response.ok(bytes);
+
+    expect(response.contentLength, 4);
+    expect(await response.read().first, <int>[1, 2, 3, 4]);
```
natebosch, line 50: I'm pretty sure the `<int>` here isn't changing behavior. We're still gonna hit the `equals` matcher which doesn't check generics on collections. If checking the type is an important behavior we'll want to use `isA`

```suggestion
    expect(await response.read().single, isA<List<int>>().having((v) => v, 'content', [1, 2, 3, 4]));
```

natebosch (commented):
Can you add a CHANGELOG entry and bump the patch version?

lib/src/body.dart
jonahwilliams: Done

lib/src/body.dart
jonahwilliams: Done

test/response_test.dart
jonahwilliams: Ahh, I was looking for that matcher but didn't find it - thank you!

test/response_test.dart
jonahwilliams: done

test/response_test.dart
jonahwilliams: done

lib/src/body.dart
```diff
@@ -52,9 +52,13 @@ class Body {
...
       }
+    } else if (body is List<int>) {
+      // Avoid performance overhead from an unnecessary cast.
+      contentLength = body.length;
+      stream = Stream.value(body);
     } else if (body is List) {
       contentLength = body.length;
-      stream = Stream.fromIterable([body.cast()]);
+      stream = Stream.value(body.cast());
     } else if (body is Stream) {
```
natebosch, line 62: I wonder, while we are here should we assume that `Stream.cast` is likely to have similar overhead?

```suggestion
    } else if (body is Stream<int>) {
      stream = body;
    } else if (body is Stream) {
```

test/response_test.dart
```diff
@@ -25,6 +26,33 @@ void main() {
...
+
+  test('Copies a dynamic list of int elements', () async {
+    var bytes = <dynamic>[1, 2, 3, 4];
+    var response = Response.ok(bytes);
+
+    expect(response.contentLength, 4);
+    expect(
+        await response.read().single,
+        isA<List<int>>()
+            .having((List<int> values) => values, 'values', [1, 2, 3, 4]));
```
natebosch, line 53:
```suggestion
            .having((values) => values, 'values', [1, 2, 3, 4]));
```

natebosch (approved):

lib/src/body.dart
jonahwilliams: Definitely, seems worth the optimization

test/response_test.dart
jonahwilliams: done

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.1.1
+
+* Avoid wraping response bodies that already contained `List<int>` or `Stream<List<int>>` in a `CastList`/`CastStream` to improve performance.
```
natebosch, line 3:
```suggestion
* Avoid wrapping response bodies that already contained `List<int>` or
  `Stream<List<int>>` in a `CastList`/`CastStream` to improve
  performance.
```

jonahwilliams:
I don't have any merge permissions btw 😄

lib/src/body.dart
```diff
@@ -52,9 +52,16 @@ class Body {
...
         stream = Stream.fromIterable([encoded]);
       }
+    } else if (body is List<int>) {
+      // Avoid performance overhead from an unnecessary cast.
+      contentLength = body.length;
+      stream = Stream.value(body);
     } else if (body is List) {
       contentLength = body.length;
-      stream = Stream.fromIterable([body.cast()]);
+      stream = Stream.value(body.cast());
```
leafpetersen, line 61: If this is likely to be iterated over multiple times, you may be better to eagerly copy the list with `List.from` instead of using `cast`.

natebosch:
This is published

lib/src/body.dart
natebosch: My hunch is this isn't a problem. I don't expect multiple iterations in most cases. Usually when we would need this cast it would be a `Response` which will be passed to `dart:io` which I expect iterates once. Otherwise, when this stream is getting passed to user code (where it may be iterated more than once) I'd expect the optimizations in this PR should kick in since it will most often be a `Stream<List<int>>` coming from `dart:io` to start with.

PR #183, opened 2021-05-10
Remove content-length for notModified responses

natebosch:
It is an error to include a non-zero content-length when there is no
body. In some places, especially those where we are implementing a
proxy, we were previously relying on behavior in shelf which would zero
out the content length when there is no body. After allowing a content
length without a body for `HEAD` requests we lost the safety for other
request types.

The errors that have surfaced since published are related to
`Response.notModified` which should never have a `content-length`, so
handling it there should retain most of the safety we had before. There
is still the possibility for problems in other response types which will
need to be handled at the usage site.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
-## 1.1.3-dev
+## 1.1.3
+
+* Automatically remove `content-length` header from a `Response.notModified`.
+  Restores some of the safety around malformed requests that was removed in
+  `1.1.2` where we started allowing `content-length` for some responses without
+  bodies.
 
 ## 1.1.2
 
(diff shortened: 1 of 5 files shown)


grouma (approved):

kevmoo:
Nit: it's totally fine for a 304 to include a `content-length` – as long as it matches what WOULD be sent

natebosch:
> Nit: it's totally fine for a 304 to include a `content-length` – as long as it matches what WOULD be sent

Is that a bug in `dart:io` in that case?

kevmoo:
That's what I'm thinking!

Or the code implemented in google3

kevmoo:
GREAT CATCH!

PR #196, opened 2021-07-01
Added `MiddlewareExtensions` which provides `addMiddleware` and `addHandler`

kevmoo:
...to `Middleware` instances.
Provides easier access to the composition capabilities offered by `Pipeline`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.2.0
+
+* Added `MiddlewareExtensions` which provides `addMiddleware` and `addHandler`
+  to `Middleware` instances. Provides easier access to the composition
+  capabilities offered by `Pipeline`.
+
 ## 1.1.4
 
 * Documentation improvements.
--- a/lib/src/middleware_extensions.dart
+++ b/lib/src/middleware_extensions.dart
@@ -1,0 +1,14 @@
+import 'handler.dart';
+import 'middleware.dart';
+
+/// Extensions on [Middleware] to aid in composing [Middleware] and [Handlers].
+///
+/// These members can be used in place of [Pipeline].
+extension MiddlewareExtensions on Middleware {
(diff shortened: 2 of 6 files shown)


lib/src/middleware_extensions.dart
```diff
@@ -0,0 +1,14 @@
+import 'handler.dart';
+import 'middleware.dart';
+
+/// Extensions on [Middleware] to aid in composing [Middleware] and [Handlers].
+///
+/// These members can be used in place of [Pipeline].
+extension MiddlewareExtensions on Middleware {
+  /// Merges `this` and [other] into a new [Middleware].
+  Middleware addMiddleware(Middleware other) =>
+      (Handler handler) => this(other(handler));
```
natebosch, line 10: Would nesting this very deeply hurt performance compared to using `Pipeline`? I'm worried about introducing a convenience method if some power users are going to have to steer away from it.

kevmoo: I think the perf will be *identical* to what exists now. Look at `Pipeline` – it's the same thing!

kevmoo: https://github.com/dart-lang/shelf/blob/a16ea0356c3f73fb9e44973b6a52a06c5546fc75/lib/src/pipeline.dart#L42

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.2.0
```
natebosch, line 1:
```suggestion
## 1.2.0-dev
```

natebosch (approved):

CHANGELOG.md
kevmoo: I was going to publish right away...

natebosch: After it's synced right...

kevmoo: Oh...yes. Of course. 🤷

Issue #202, opened 2021-09-28
Suggestion: always run serve_requests in `runZoneGuarded`


annagrin:
`serveRequests` does not always wrap listening to http requests in a `runZoneGuarded` wrap:

https://github.com/dart-lang/shelf/blob/6e38575985a8094f4ac172228a6711196d9ae91e/lib/shelf_io.dart#L73
https://github.com/dart-lang/shelf/blob/6e38575985a8094f4ac172228a6711196d9ae91e/lib/src/util.dart#L16

This leads to hard to debug async exceptions later if a something up its call chain adds its own `runZoneGuarded` wrapper - in that case exceptions stop from being caught and ignored inside `serveRequests` and fall to the next root zone exception hander, which unexpectedly changes to logic of the calling code. Note that adding such top level handler will change behavior of all `serveRequests` call, including the ones coming from imported libraries, which makes fixing those new issues particularly complicated.

Can we always run code inside `serveRequests` in a `runZoneGuarded`, and provide it with an error handler instead for that cases where error behavior needs to be controlled by the user?

kevmoo:
I think I buy that. Thoughts @natebosch @jakemac53 @lrhn ?

kevmoo:
@natebosch looked at pulling out the error handling to do something different during the null safety push and we never got anywhere.

Might be worth revisiting, though...

jakemac53:
Do we know what the rationale for the existing behavior was? It does seem problematic to me, but it also seems like somebody went to a fair bit of effort to make it behave that way?

kevmoo:
@nex3 had good reasons, I'm sure – perhaps she can shed some light?

natebosch:
Yeah the current behavior is magic and confusing - the hidden interactions with error zones does cause problems.

https://github.com/dart-lang/shelf/issues/2#issuecomment-742081585

My preference is to remove anything related to error zones in this package, but we'd need to be careful rolling it out.

kevmoo:
The OTHER idea.

Chatting with a community member about this.

Deprecate `shelf_io.dart` and create another, more opinionated package to handle the "wire up the server" bits.

annagrin:
I am good with anything that does not change behavior if a top level handler is added, we would need to update a bunch of existing libraries in that case though. My current case - ddr has a top level handler that has been added to collect crash metrics, but it runs a bunch of http servers via libraries such as dds, dwds, devtools, build daemon etc.

nex3:
[...]

PR #204, opened 2021-10-05
Add serveGuarded and serveRequestsGuarded API

annagrin:
`serveRequests` does not always wrap listening to http requests in a `runZoneGuarded` wrap.

This leads to hard to debug async exceptions later if something up its call chain adds its own
 `runZoneGuarded` wrapper - in that case exceptions stop being caught and ignored inside
 `serveRequests` and fall to the next root zone exception handler, which unexpectedly changes
the logic of the calling code.

Note that adding such top level handler will change behavior of all `serveRequests` calls,
including the ones coming from imported libraries, which makes fixing those new issues
particularly complicated.

This change adds two new APIs, `serveGuarded` and `serveRequestsGuarded`, that alleviate
the issue by always serving requests in an error zone and enabling the user to provide
 an error handler. A simple print-and-continue error handler is used by default.

In addition, updated the message on the default error handler to reflect source of the
errors.

Closes: https://github.com/dart-lang/shelf/issues/202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.2.1-dev
+## 1.3.0-dev
 
 * Added `serveGuarded` and `serveRequestGuarded` API that always serve
   requests in an error zone, capture async error traces, and use provided
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 1.2.1-dev
 
+* Added `serveGuarded` and `serveRequestGuarded` API that always serve
+  requests in an error zone, capture async error traces, and use provided
+  error handler to handle errors.
+
 ## 1.2.0
 
 * Added `MiddlewareExtensions` which provides `addMiddleware` and `addHandler`
--- a/lib/shelf_io.dart
+++ b/lib/shelf_io.dart
@@ -36,6 +36,52 @@ export 'src/io_server.dart' show IOServer;
 ///
 /// If a [securityContext] is provided an HTTPS server will be started.
 ///
+/// Serves requests in an error zone, using [onError] to handle errors, if
+/// provided. Uses a default error handler that prints error summary if
+/// [onError] is not provided.
+///
+/// See the documentation for [HttpServer.bind] and [HttpServer.bindSecure]
+/// for more details on [address], [port], [backlog], and [shared].
+Future<HttpServer> serveGuarded(
+  Handler handler,
+  address,
+  int port, {
+  SecurityContext? securityContext,
+  int? backlog,
+  bool shared = false,
+  void Function(Object, StackTrace) onError = _logServeRequestErrors,
+}) async {
+  backlog ??= 0;
+  var server = await (securityContext == null
+      ? HttpServer.bind(address, port, backlog: backlog, shared: shared)
+      : HttpServer.bindSecure(
+          address,
+          port,
+          securityContext,
+          backlog: backlog,
+          shared: shared,
+        ));
+  serveRequestsGuarded(server, handler, onError: onError);
+  return server;
+}
+
+/// Serve a [Stream] of [HttpRequest]s with error handling.
+///
+/// Serves requests in an error zone, using [onError] to handle errors, if
+/// provided. Uses a default error handler that prints error summary if
+/// [onError] is not provided.
+void serveRequestsGuarded(Stream<HttpRequest> requests, Handler handler,
+    {void Function(Object, StackTrace) onError = _logServeRequestErrors}) {
+  return Chain.capture(() {
+    serveRequests(requests, handler);
+  }, onError: onError);
+}
+
+/// Starts an [HttpServer] that listens on the specified [address] and
+/// [port] and sends requests to [handler].
+///
+/// If a [securityContext] is provided an HTTPS server will be started.
+///
 /// See the documentation for [HttpServer.bind] and [HttpServer.bindSecure]
 /// for more details on [address], [port], [backlog], and [shared].
 Future<HttpServer> serve(
@@ -72,9 +118,7 @@ Future<HttpServer> serve(
 void serveRequests(Stream<HttpRequest> requests, Handler handler) {
   catchTopLevelErrors(() {
     requests.listen((request) => handleRequest(request, handler));
-  }, (error, stackTrace) {
-    _logTopLevelError('Asynchronous error\n$error', stackTrace);
-  });
+  }, _logServeRequestErrors);
 }
 
 /// Uses [handler] to handle [request].
@@ -254,3 +298,6 @@ void _logTopLevelError(String message, StackTrace stackTrace) {
   stderr.writeln(message);
   stderr.writeln(chain);
 }
+
+void _logServeRequestErrors(Object error, StackTrace stackTrace) =>
+    _logTopLevelError('Serve request error:\n$error', stackTrace);
--- a/lib/shelf_io.dart
+++ b/lib/shelf_io.dart
@@ -31,50 +31,39 @@ import 'src/util.dart';
 
 export 'src/io_server.dart' show IOServer;
 
-/// Starts an [HttpServer] that listens on the specified [address] and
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 1.2.1-dev
 
+* Added `serveGuarded` and `serveRequestGuarded` API that always serve
```
kevmoo, line 3: Should make this 1.3.0(-dev) – since we're adding API

kevmoo (commented):
Worth at least a couple of tests?

lib/shelf_io.dart
```diff
@@ -31,6 +31,52 @@ import 'src/util.dart';
...
+}
+
+/// Serve a [Stream] of [HttpRequest]s with error handling.
+///
+/// Serves requests in an error zone, using [onError] to handle errors, if
+/// provided. Uses a default error handler that prints error summary if
+/// [onError] is not provided.
+void serveRequestsGuarded(Stream<HttpRequest> requests, Handler handler,
+    {void Function(Object, StackTrace) onError = _logServeRequestErrors}) {
+  return Chain.capture(() {
```
natebosch, line 75: I don't think we want to use `Chain.capture` for this since it comes with overhead.

kevmoo:
To pile on: wonder if it would be good to give references to these functions from the existing ones – so folks know there are other options?

lib/shelf_io.dart
```diff
@@ -31,6 +31,52 @@ import 'src/util.dart';
...
+/// provided. Uses a default error handler that prints error summary if
+/// [onError] is not provided.
+///
+/// See the documentation for [HttpServer.bind] and [HttpServer.bindSecure]
+/// for more details on [address], [port], [backlog], and [shared].
+Future<HttpServer> serveGuarded(
+  Handler handler,
+  address,
+  int port, {
+  SecurityContext? securityContext,
```
natebosch, line 49: If we are moving to a whole new API anyway we should spend some time investigating whether the design is the right one for current usage.

My gut is that wrapping the `HttpServer.bind` call and forwarding arguments like this isn't adding much value, and that the `serveRequests` API is readable enough if call sites need to inline the `HttpServer.bind`.

lib/shelf_io.dart
```diff
@@ -31,6 +31,52 @@ import 'src/util.dart';
...
+          shared: shared,
+        ));
+  serveRequestsGuarded(server, handler, onError: onError);
+  return server;
+}
+
+/// Serve a [Stream] of [HttpRequest]s with error handling.
+///
+/// Serves requests in an error zone, using [onError] to handle errors, if
+/// provided. Uses a default error handler that prints error summary if
```
natebosch, line 71: We should consider making the handler mandatory. We could give an example usage that would match the default behavior, but I think we expect all use cases to want to do something better than the default so that should be reflected in the API.

natebosch:
> wonder if it would be good to give references to these functions from the existing ones – so folks know there are other options?

I'd like to standardize onto the new API and drop the old one.

lib/shelf_io.dart
annagrin: Interesting, I can add a flag for skipping the capture. Unfortunately traces are not very useful without it - how much is the overhead?

lib/shelf_io.dart
annagrin: > My gut is that wrapping the HttpServer.bind call and forwarding arguments like this isn't adding much value, and that the serveRequests API is readable enough if call sites need to inline the HttpServer.bind.

I am not really following - do you mean that `serve` and `serveGuarded` are not useful? I haven's seen much use of `serve` myself, maybe they both need to be removed. I only added a guarded version of the existing one.

lib/shelf_io.dart
annagrin: Makes sense, the default behavior is really confusing - the errors will be printing to the console with not much context.

CHANGELOG.md
annagrin: sure:) I wanted to get the API right first...

lib/shelf_io.dart
annagrin: I think you are right and `serveXXX` are not very useful - normally users would have modifications to the starting the server code, be able to access the server (for stopping) and error handling in case it didn't start properly.

annagrin: Btw would you like me to just remove serve and serveRequests? Or do we need to deprecate them?

annagrin:
>> wonder if it would be good to give references to these functions from the existing ones – so folks know there are other options?

> I'd like to standardize onto the new API and drop the old one.

Should I remove the old API in this PR, or is there a deprecation procedure?

jakemac53:
> Should I remove the old API in this PR, or is there a deprecation procedure?

You can add an `@Deprecated('Use serveGuarded') annotation`.

lib/shelf_io.dart
```diff
@@ -31,6 +31,52 @@ import 'src/util.dart';
...
+  Handler handler,
+  address,
+  int port, {
+  SecurityContext? securityContext,
+  int? backlog,
+  bool shared = false,
+  void Function(Object, StackTrace) onError = _logServeRequestErrors,
+}) async {
+  backlog ??= 0;
+  var server = await (securityContext == null
```
jakemac53, line 55: This can also fail - it should be ran in an error handling zone as well

natebosch: This error would surface in the returned `Future` instead of disappear as uncaught async, so I think it's OK to let it skip the `onError` callback. Although I also think we should consider dropping the method entirely and push the `HttpServer.bind` call into call-site - but we should look at more usage to figure out if that is sensible.

lib/shelf_io.dart
natebosch: I think we'll need to deprecate them. We might consider having a bit of time before we add the annotation so we can get some migrations out of the way ahead of CI turning red.

I've looked at this a little bit closer and I think I'm more convinced that we want to leave the `HttpServer` stuff to the clients. The caller already needs to be worried about calling `close()` on that server, so it makes sense to me to push the entirety of lifetime control for the server to the caller.

@kevmoo @jakemac53 - what do you think of trying something with an entirely different API?

```dart
extension ServeRequests on Handler {
  void serveRequests(Stream<HttpRequest> requests, {required void Function(Object, StackTrace) onError}) {
    runZonedGuarded(() {
      requests.forEach((request) => handleRequest(request, this));
    }, onError);
  }
}
```

lib/src/io_server.dart
```diff
@@ -52,7 +52,9 @@ class IOServer implements Server {
     }
     _mounted = true;
 
-    serveRequests(server, handler);
+    serveRequestsGuarded(server, handler, (e, s) {
```
annagrin, line 55: I think we should modify the `mount` API here to take an error handler

annagrin:
@natebosch @kevmoo @jakemac53 I think this PR is going towards revisiting some API and looking at their usages, I would be happy if one of you decides to tae over:)

jakemac53:
I am fine with just doing the workarounds for now where we need and doing a more general rewrite here.

kevmoo:
@annagrin – any plans here?

natebosch:
I had started thinking about the design here and we decided to hold off on this PR, but then I never finalized a plan. I had some work towards a cleaner version of `serveRequests`... I'll see if I can track that down.

annagrin:
@kevmoo I will defer the proper fix to the active library developers (@natebosch I presume). To fix the issues caused by this, I had to wrap the `serveRequests` call in an error zone in all affected places (there are probably more, but the ones I found were in dwds, webdev, devtools server, ddr, and if I remember correctly build_daemon). The `SocketException` crash rate dropped 10x but they are still happening to approx 1.5% of our sessions. I might have missed some other cases, which may or may not relate to this issue.

mosuem:
Closing this, as it is stale and the files are out of date. Feel free to reopen in case you still want to land it!

PR #208, opened 2021-10-09
Multipart support

### LINKED ISSUE #119: Provide support for multipart requests
Currently, there's no way to get the files from HTTP requests with [the multipart content-type](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). This makes it impossible to implement HTTP servers that support file uploads, or to test a client that makes multipart requests.

simolus3:
Adds extensions on `Request` to handle multipart data.

Closes #119.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
-## 1.2.1-dev
+## 1.3.0-dev
 
+* Add the `ReadMultipartRequest` extension on `Request` to handle multipart requests:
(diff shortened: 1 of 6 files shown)


kevmoo:
Nice – although I wonder if this should be a separate package. Just trying to keep things small here.

@jakemac53 ? @natebosch ?

simolus3:
Fair enough, but then we should probably close the issue as wontfix as well?

kevmoo:
@simolus3 – or we wait until you publish your package and we link it?

PR #230, opened 2022-05-10
Add examples

ushieru:
Example files in isolation and with shelf_router.


pkgs/shelf_web_socket/example/example_shelf_router.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:shelf/shelf.dart';
+import 'package:shelf/shelf_io.dart' as io;
+import 'package:shelf_router/shelf_router.dart';
+import 'package:shelf_web_socket/shelf_web_socket.dart';
+
```
kevmoo, line 5: We likely don't want to add dependencies to `shelf_router` here!

PR #242, opened 2022-06-15
prep for publishing the shelf packages

devoncarew:
Prep for publishing the shelf packages:
- rev all the versions (and check that they are the correct patch versions from last published)
- make sure there is content in the latest changelog entries
- some minor normalization of the most recent changelog entries


kevmoo (approved):
That's a lot of packages to publish for not a lot of user benefit 🤷

PR #246, opened 2022-06-27
Exposing Request object in logger

busslina:


pkgs/shelf/lib/src/middleware/logger.dart
```diff
@@ -19,7 +20,8 @@ import '../middleware.dart';
 /// The `isError` parameter indicates whether the message is caused by an error.
 ///
 /// If [logger] is not passed, the message is just passed to [print].
-Middleware logRequests({void Function(String message, bool isError)? logger}) =>
+Middleware logRequests(
```
kevmoo, line 23: I see where you're going here, but this is a BIG breaking change.

busslina: Why is it a big breaking change? Migrating from old to new code by adding one more argument.

natebosch: What is the motivation for this change? What else are you looking for that isn't in the `message`? Do you have an example logger callback you are planning to use with this signature?

busslina: My goal is to access some data stored on the Request context map and print info with the logger.

natebosch:
Closing since I this is not a signature change we plan on making.

PR #263, opened 2022-08-11
adapter for flutter web or dart web

yk3372:
In web's cross domain request, serverRequest.method is OPTIONS.
In Dart Server, we need to use the truly request from user.
And this truly method is come from http-header:'Access-Control-Request-Method', such as POST/GET etc.

The doc is here : https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

My PR doesn't effect normal request, no cross domain request，it will use serverRequest.method. Please Review it, thanks.
@kevmoo @natebosch

### DIFF
--- a/pkgs/shelf_proxy/lib/shelf_proxy.dart
+++ b/pkgs/shelf_proxy/lib/shelf_proxy.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:io';
 
 import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as p;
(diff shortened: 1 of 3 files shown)


pkgs/shelf_proxy/lib/shelf_proxy.dart
```diff
@@ -40,7 +41,15 @@ Handler proxyHandler(Object url, {http.Client? client, String? proxyName}) {
     // TODO(nweiz): Handle TRACE requests correctly. See
     // http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.8
     final requestUrl = uri.resolve(serverRequest.url.toString());
-    final clientRequest = http.StreamedRequest(serverRequest.method, requestUrl)
+
```
kevmoo, line 44: We should have a changelog entry for this. It'd also be good to reference a spec/standard about this change. I'm not familiar.

ALSO: there should be a test!

yk3372: I don't how to write test, but I can support test demo.
In debug mode, web site want fetch http://targetHost/show/upgrade.
so, I use proxy fetch http://localhost:8888/show/upgrade.
And dart server is localhost:8888.

The dart server need to fetch the truly url: http://targetHost/show/upgrade.
But serverRequest.method is OPTIONS, this method can't get data, and request fail.
The dart server need use the truly method from http-header:'Access-Control-Request-Method', is GET.

Below is the screenshot:
image
image

@kevmoo Can you help me to write the test case? Thanks

kevmoo: Okay. Looking here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Request-Method – it's unclear that the correct behavior is two change the method from the original request, though.

This seems like a specific need of your scenario, not something that should be done generally for every projy usage...

natebosch: > This seems like a specific need of your scenario, not something that should be done generally for every proxy usage...

+1

This does not look to match the behavior I would expect for a proxy. I don't understand how the real server could respond correctly to an `OPTIONS` request if it does not see that request type.

PR #272, opened 2022-08-24
Add a poweredByHeader argument to server

### LINKED ISSUE #270: The Server header cannot be removed (docs suggest it shouldn't even be added by default)
Starting up a hello world server with shelf I get this response:

```
HTTP/1.1 200 OK
date: Sat, 20 Aug 2022 13:15:00 GMT
content-length: 13
x-frame-options: SAMEORIGIN
content-type: text/plain; charset=utf-8
x-xss-protection: 1; mode=block
x-content-type-options: nosniff
server: dart:io with Shelf

Hello World!
```

Notice the `server` header. I don't need most of those headers, so I am doing this to remove them:

```dart
  var server =
      await shelf_io.serve(handler, config.address, config.port, shared: true);

  server.defaultResponseHeaders
    ..removeAll('x-frame-options')
    ..removeAll('x-xss-protection')
    ..removeAll('x-content-type-options')
    ..removeAll('server');
  server.serverHeader = null;
```

All headers I removed are gone, except for `Server`:

```
HTTP/1.1 200 OK
content-type: text/plain; charset=utf-8
date: Sat, 20 Aug 2022 13:24:08 GMT
server: dart:io with Shelf
content-length: 13

Hello World!
```

According to the docs, the `Server` header shouldn't even be there....

From `/dart/2.17.6/libexec/lib/_http/http.dart:124`:

```dart
  /// If the `Server` header is added here and the `serverHeader` is set as
  /// well then the value of `serverHeader` takes precedence.
  HttpHeaders get defaultResponseHeaders;
```

From `/usr/local/Cellar/dart/2.17.6/libexec/lib/_http/http.dart:111`:

```dart
  /// If [serverHeader] is `null`, no `Server` header will be added to each
  /// response.
  ///
  /// The default value is `null`.
  String? serverHeader;
```

Is there another way to remove that? Are the docs wrong?

natebosch:
Fixes #270

Allow passing `null` to omit the header.


kevmoo (approved):
Love it! Well done!

Worth a version bump and a changelog entry!

pkgs/shelf/lib/shelf_io.dart
```diff
@@ -57,6 +57,8 @@ Future<HttpServer> serve(
           backlog: backlog,
           shared: shared,
         ));
+  // See https://webtechsurvey.com/response-header/x-powered-by
+  server.defaultResponseHeaders.add('x-powered-by', 'Dart with package:shelf');
```
kevmoo, line 61: Hrm. I wonder if we should check to see if this is already added and skip?
Should this be a first writer win or last writer win thing? 🤷

natebosch: It can't have been added at this point - we created the `HttpClient` above.

In a refactoring I had planned though, we don't create an `HttpClient` though, and instead take a `Stream<HttpRequest>` to allow the caller to create the client. That design wouldn't be compatible with this approach, the caller would need to set the default themselves.

natebosch: Err, `HttpServer`

kevmoo: ack...

natebosch:
Hmm, this doesn't work for the `serveRequests` code path.

kevmoo:
Right...grr...

pkgs/shelf/lib/shelf_io.dart
```diff
@@ -70,9 +78,16 @@ Future<HttpServer> serve(
 /// console and cause a 500 response with no body. Errors thrown asynchronously
 /// by [handler] will be printed to the console or, if there's an active error
 /// zone, passed to that zone.
-void serveRequests(Stream<HttpRequest> requests, Handler handler) {
+///
+/// {@macro shelf_io_header_defaults}
```
kevmoo, line 82: Aren't macros NICE!!

kevmoo (approved):

pkgs/shelf/lib/shelf_io.dart
```diff
@@ -39,13 +39,21 @@ export 'src/io_server.dart' show IOServer;
 ///
 /// See the documentation for [HttpServer.bind] and [HttpServer.bindSecure]
 /// for more details on [address], [port], [backlog], and [shared].
+///
+/// {@template shelf_io_header_defaults}
+/// Every response will get a "date" header and an "X-Powered-By" header.
+/// If the either header is present, it will not be overwritten.
+/// Pass [poweredByHeader] to set the default content for "X-Powered-By", pass
+/// null to omit this header..
```
kevmoo, line 47:
```suggestion
/// `null` to omit this header..
```

pkgs/shelf/lib/shelf_io.dart
```diff
@@ -39,13 +39,21 @@ export 'src/io_server.dart' show IOServer;
 ///
 /// See the documentation for [HttpServer.bind] and [HttpServer.bindSecure]
 /// for more details on [address], [port], [backlog], and [shared].
+///
+/// {@template shelf_io_header_defaults}
+/// Every response will get a "date" header and an "X-Powered-By" header.
+/// If the either header is present, it will not be overwritten.
```
kevmoo, line 45:
```suggestion
/// If the either header is present in the `Response`, it will not be overwritten.
```

maybe?

natebosch: Works for me.

PR #280, opened 2022-08-30
Unify on one analysis_options file

kevmoo:
Apply fixes to shelf_router and shelf_router_generator


pkgs/shelf_router/lib/src/route.dart
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import 'package:meta/meta.dart' show sealed;
-import 'package:shelf_router/src/router.dart';
+import 'router.dart';
```
devoncarew, line 16: organize imports?

pkgs/shelf_router/lib/src/router.dart
```diff
@@ -294,7 +295,7 @@ class _RouteNotFoundResponse extends Response {
   Response change({
     Map<String, /* String | List<String> */ Object?>? headers,
     Map<String, Object?>? context,
-    body,
+    Object? body,
```
devoncarew, line 298: 👍

devoncarew (approved):

pkgs/shelf_router/lib/src/route.dart
kevmoo: prefer relative imports...

devoncarew: For sure, but this changed the import from a package to a file import, and the 'organize import' command would insert a blank line between these two imports now :)

kevmoo: ah! Hrm...yup

PR #301, opened 2022-10-22
Properly check for callback arity in `shelf_web_socket`

simolus3:
Pretty sure this is the right way to do this now since `void Function(Null, Null)` is no longer the topmost type for binary functions.

I really don't understand why tests don't catch this, but this fixes issues surfacing in client packages (in particular, https://github.com/dart-lang/build/issues/3379).

### DIFF
--- a/pkgs/shelf_web_socket/test/web_socket_test.dart
+++ b/pkgs/shelf_web_socket/test/web_socket_test.dart
@@ -2,14 +2,13 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: inference_failure_on_untyped_parameter
-
 import 'dart:io';
 
 import 'package:http/http.dart' as http;
 import 'package:shelf/shelf_io.dart' as shelf_io;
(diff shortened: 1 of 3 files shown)


kevmoo:
Thoughts on this @natebosch ? @devoncarew ?

kevmoo:
it'd be kinda nice if there was a test to validate this fix! 😁

pkgs/shelf_web_socket/test/web_socket_test.dart
```diff
@@ -2,14 +2,13 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: inference_failure_on_untyped_parameter
```
kevmoo, line 5: 👍

natebosch:
Our tests did not catch this, because `Function(Null, Null)` _is_ a subtype of `Function(dynamic, dynamic)`, which is what we had in the tests. Most production code would use type arguments since inference doesn't help. Adding type arguments to our test callbacks, which better match production code, catches this bug.

natebosch (approved):
Thanks for the fix!

PR #379, opened 2023-09-16
Added support for WebSockets

Martoxdlol:
Hello!
I added support for proxying websockets. It is implemented checking the 'Upgrade' header and using `shelf_web_socket` and `web_socket_channel` packages.

### DIFF
--- a/pkgs/shelf_proxy/lib/shelf_proxy.dart
+++ b/pkgs/shelf_proxy/lib/shelf_proxy.dart
@@ -7,6 +7,9 @@ import 'dart:async';
 import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as p;
 import 'package:shelf/shelf.dart';
+import 'package:shelf_web_socket/shelf_web_socket.dart';
+import 'package:web_socket_channel/io.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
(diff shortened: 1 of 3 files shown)


Martoxdlol:
Hello, could someone review this?

kevmoo:
@natebosch @devoncarew @brianquinlan – thoughts?

natebosch:
This package currently does not have imports to `dart:io`. I don't know that there is utility in running this code anywhere without `dart:io`, but adding the new dependencies and the new platform restriction is probably enough reason to not add this support.

I wonder if there is some way can make this pluggable.

Or is the `import 'package:web_socket_channel/io.dart';` required - maybe we can use the cross platform import for connecting the socket?

Martoxdlol:
> This package currently does not have imports to `dart:io`. I don't know that there is utility in running this code anywhere without `dart:io`, but adding the new dependencies and the new platform restriction is probably enough reason to not add this support.
> 
> I wonder if there is some way can make this pluggable.
> 
> Or is the `import 'package:web_socket_channel/io.dart';` required - maybe we can use the cross platform import for connecting the socket?

Hmmm I didn't think about that. I will look for another solution. 

Maybe I can make a separate package that adds websockets or something.

PR #395, opened 2023-12-05
Add context to `Response` for including `file` and `file_not_found` to be identified by other `shelf` `Middleware`.

gmpassos:
- [x ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

---
Updated:

With the `Response.context` populated with the processed `File`, other `Middleware` can manipulate the `Response` in an optimized way and also ensure the correct processed `File`. Without that, another `Middleware` would need to guess the processed `File`, and in some cases, re-resolve `File.stat` and also fix the directory path to the "index.html" path.

### DIFF
--- a/pkgs/shelf_static/lib/src/static_handler.dart
+++ b/pkgs/shelf_static/lib/src/static_handler.dart
@@ -82,7 +82,10 @@ Handler createStaticHandler(String fileSystemPath,
     }
 
     if (fileFound == null) {
-      return Response.notFound('Not Found');
+      return Response.notFound(
+        'Not Found',
+        context: _buildResponseContext(fileNotFound: fileFound),
+      );
     }
     final file = fileFound;
 
@@ -117,6 +120,18 @@ Handler createStaticHandler(String fileSystemPath,
         return mimeResolver.lookup(file.path);
       }
     });
+  };
+}
+
+Map<String, Object>? _buildResponseContext({File? file, File? fileNotFound}) {
+  if (file == null && fileNotFound == null) return null;
+
+  // Ensure other Shelf `Middleware` can identify
+  // the processed file in the `Response` by
+  // including `file` and `file_not_found` in the context:
+  return {
+    if (file != null) 'shelf_static:file': file,
+    if (fileNotFound != null) 'shelf_static:file_not_found': fileNotFound,
   };
 }
 
@@ -184,7 +199,9 @@ Future<Response> _handleFile(Request request, File file,
   if (ifModifiedSince != null) {
     final fileChangeAtSecResolution = toSecondResolution(stat.modified);
     if (!fileChangeAtSecResolution.isAfter(ifModifiedSince)) {
-      return Response.notModified();
+      return Response.notModified(
+        context: _buildResponseContext(file: file),
+      );
     }
   }
 
@@ -199,6 +216,7 @@ Future<Response> _handleFile(Request request, File file,
       Response.ok(
         request.method == 'HEAD' ? null : file.openRead(),
         headers: headers..[HttpHeaders.contentLengthHeader] = '${stat.size}',
+        context: _buildResponseContext(file: file),
       );
 }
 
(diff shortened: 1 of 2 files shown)


kevmoo:
i see where you're going here.

I'd want to chat with @natebosch or @devoncarew on their thoughts

This change should be documented and tested, too!

gmpassos:
With the Response.context populated with the processed File, other Middleware can manipulate the Response in an optimized way and also ensure the correct processed File. Without that, another Middleware would need to guess the processed File, and in some cases, re-resolve File.stat and also fix the directory path to the "index.html" path.

pkgs/shelf_static/lib/src/static_handler.dart
```diff
@@ -39,6 +39,11 @@ final _defaultMimeTypeResolver = MimeTypeResolver();
 ///
 /// Specify a custom [contentTypeResolver] to customize automatic content type
 /// detection.
+///
+/// The [Response.context] will be populated with "shelf_static:file" or
+/// "shelf_static:file_not_found" with the resolved [File] for the [Response],
+/// while respecting [serveFilesOutsidePath] to prevent exposing a [File] path
```
natebosch, line 45: Call out the specific behavior for `serveFilesOutsidePath`. Maybe something like

"If the file is considered not found because it is outside of the [fileSystemPath] and [serveFilesOutsidePath] is false then neither key will be included in the context."

gmpassos: How about?:

```markdown
/// The [Response.context] will be populated with "shelf_static:file" or
/// "shelf_static:file_not_found" with the resolved [File] for the [Response].
/// If the file is considered not found because it is outside of the
/// [fileSystemPath] and [serveFilesOutsidePath] is false, then neither key
/// will be included in the context.
```

- If you confirm this version I will change the text in the code.

natebosch: > How about?:

Looks Good To Me

gmpassos:
What's the next step? Regards.

kevmoo:
need changelog entry and tests

natebosch:
Is this still something you're interested in landing - can you take a look at adding tests?

gmpassos:
This is important. 

I will try to add some tests, but I don't have time for that in the next 15 days.

natebosch:
Can you give an example of how you plan to use this? Did you consider any alternative patterns, such as always storing the key `shelf_static:file`, and storing a separate boolean field for whether it was found?

Also cc @brianquinlan - can you think of any risks in keeping the `File` object in the heap for longer than we would have previously.

gmpassos:
> Can you give an example of how you plan to use this? Did you consider any alternative patterns, such as always storing the key `shelf_static:file`, and storing a separate boolean field for whether it was found?

You can take a look at real-world code that is waiting for this:
https://github.com/Colossus-Services/bones_api/blob/master/lib/src/bones_api_server.dart#L1788

Note that if `'shelf_static:file'` and `'shelf_static:file_not_found'` are not present, we need to recalculate the `File` path, validate it, and ensure (again) that it's inside the "root" directory (simulating the same `shelf:createStaticHandler` logic).

The `context['file']` is then used by the cache "middleware," which will use the `File` instance to validate the cached files and dispose of the cached data if the `File` changes:
https://github.com/Colossus-Services/bones_api/blob/master/lib/src/bones_api_server_cache.dart#L351

brianquinlan:
### 

> Can you give an example of how you plan to use this? Did you consider any alternative patterns, such as always storing the key `shelf_static:file`, and storing a separate boolean field for whether it was found?
> 
> Also cc @brianquinlan - can you think of any risks in keeping the `File` object in the heap for longer than we would have previously.

A `File` object doesn't represent any native resources so keeping it alive shouldn't do anything except cost a bit of memory.

gmpassos:
@natebosch @brianquinlan Next step?

gmpassos:
This PR will continue at PR
https://github.com/dart-lang/shelf/pull/436

PR #398, opened 2023-12-13
Run web tests with wasm with dev Dart sdk

kevmoo:


pkgs/shelf_test_handler/mono_pkg.yaml
```diff
@@ -17,3 +17,5 @@ stages:
     os:
     - linux
     - windows
+  - test: --test-randomize-ordering-seed=random -p chrome -c dart2wasm
```
jakemac53, line 20: Should we explicitly run this on linux and windows?

jakemac53 (approved):

pkgs/shelf_test_handler/mono_pkg.yaml
kevmoo: This just defaults to linux IIUC, which is fine.

PR #437, opened 2024-06-17
`Headers.fromEntries`: using `CaseInsensitiveMap.fromEntries`

gmpassos:
* `Headers`:
  * Added the constructor `Headers.fromEntries`.
  * Added the internal constructor `_fromEntries`, which uses `CaseInsensitiveMap.fromEntries` for optimization.
  * Renamed the internal constructor `_` to `_from`, redirecting it to `this._fromEntries`.

---

- [✅ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/pkgs/shelf/lib/src/headers.dart
+++ b/pkgs/shelf/lib/src/headers.dart
@@ -24,7 +24,7 @@ class Headers extends UnmodifiableMapView<String, List<String>> {
     } else if (values is Headers) {
       return values;
     } else {
-      return Headers._fromEntries(values.entries);
+      return Headers._(values.entries);
     }
   }
 
@@ -33,11 +33,11 @@ class Headers extends UnmodifiableMapView<String, List<String>> {
     if (values == null || (values is List && values.isEmpty)) {
       return _emptyHeaders;
     } else {
-      return Headers._fromEntries(values);
+      return Headers._(values);
     }
   }
 
-  Headers._fromEntries(Iterable<MapEntry<String, List<String>>> entries)
+  Headers._(Iterable<MapEntry<String, List<String>>> entries)
       : super(CaseInsensitiveMap.fromEntries(entries
             .where((e) => e.value.isNotEmpty)
             .map((e) => MapEntry(e.key, List.unmodifiable(e.value)))));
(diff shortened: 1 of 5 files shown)


gmpassos:
FYI: @kevmoo

kevmoo (approved):
I made some tweaks.

kevmoo:
@gmpassos – and some more tweaks 😄

pkgs/shelf/lib/src/headers.dart
```diff
@@ -23,15 +23,24 @@ class Headers extends UnmodifiableMapView<String, List<String>> {
       return _emptyHeaders;
     } else if (values is Headers) {
       return values;
+    } else {
+      return Headers._(values.entries);
+    }
+  }
+
+  factory Headers.fromEntries(
+      Iterable<MapEntry<String, List<String>>>? values) {
```
gmpassos, line 32: Maybe rename to `entries`?

kevmoo: done

kevmoo: yep

kevmoo:
@gmpassos – do you have benchmarks showing this makes things notably faster?

gmpassos:
> @gmpassos – we need to roll this through our internal "stuff" before we publish. To make sure nothing breaks. I'll let you know when publish happens!

Thanks for the quick response and attention. I will wait for the publication...

kevmoo:
Looks like @devoncarew landed in the Dart SDK 50 minutes ago. We'll have to wait a few hours to make sure that rolls into our internal...things...before we publish

gmpassos:
> @gmpassos – do you have benchmarks showing this makes things notably faster?

In my benchmark, I get a 1-2% improvement, but note that you need to have real-world headers (real browsers send many more headers than simple HTTP clients). Additionally, you need a heavy load, as this also affects the GC. This is more significant for slower/smaller servers. On a medium-sized server without low traffic, this won't have much impact.

I'm also looking for a way to avoid computation of all canonicalized keys just to build the headers. This will significantly improve performance, or we could use a specialized cache of canonicalized keys, but this will be a future optimization.

kevmoo:
shelf published w/ this! @gmpassos !

PR #467, opened 2025-01-07
Publish shelf_router_generator 1.1.1

johnpryan:
I think it's OK to publish this version now.

cc/ @devoncarew @kevmoo

### DIFF
--- a/pkgs/shelf_router_generator/CHANGELOG.md
+++ b/pkgs/shelf_router_generator/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.1.1-wip
+## 1.1.1
 
 * Support the latest `package:analyzer` and `package:source_gen`
 * Require `sdk: ^3.3.0`
--- a/pkgs/shelf_router_generator/pubspec.yaml
+++ b/pkgs/shelf_router_generator/pubspec.yaml
@@ -1,5 +1,5 @@
 name: shelf_router_generator
-version: 1.1.1-wip
+version: 1.1.1
 description: >
   A package:build-compatible builder for generating request routers for the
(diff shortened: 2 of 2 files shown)


devoncarew:
We'll need to rev the changelog version as well:

> error: pubspec version (1.1.1) and changelog (1.1.1-wip) don't agree

devoncarew (approved):

devoncarew:
It looks like the publish task is failing on a separate package:

```
  Package validation found the following potential issue:
  * The name of "lib/builder.dart", "builder", should match the name of the package, "shelf_router_generator".
    This helps users know what library to import.
```

I'll file an issue to address this.

johnpryan:
Can we merge without fixing the `builder.dart` filename?

kevmoo:
> It looks like the publish task is failing on a separate package:
> 
> ```
>   Package validation found the following potential issue:
>   * The name of "lib/builder.dart", "builder", should match the name of the package, "shelf_router_generator".
>     This helps users know what library to import.
> ```
> 
> I'll file an issue to address this.

This is a convention for builders we've had FOREVER. CC @jakemac53 

not sure we want to teach pub about it, though....

jakemac53:
> This is a convention for builders we've had FOREVER. CC @jakemac53
> 
> not sure we want to teach pub about it, though....

My unpopular opinion is this is a horrible recommendation period 🤣 . We should be doing the opposite and encouraging packages to have many different imports instead of one huge one.

devoncarew:
I opened https://github.com/dart-lang/pub/issues/4484 to remove or gain the ability to ignore this warning.

PR #498, opened 2026-01-20
Update test command from 'pub run test' to 'dart test'

sigurdm:
Re https://github.com/dart-lang/pub/issues/4737


kevmoo (approved):
Wow, that's old! 😄

PR #504, opened 2026-03-03
Enhance routing engine

redfieldchristabel:
### Description
This PR introduces a significant architectural enhancement to `shelf_router` by replacing the existing linear regex-based matching engine with a high-performance **Trie-based routing engine**.

#### Why this change?
The current implementation utilizes a list of regex patterns, resulting in **$O(N)$** lookup complexity where performance degrades linearly as the number of routes increases. This enhancement shifts the complexity to **$O(L)$**, where $L$ is the depth of the path segments, ensuring that routing time remains constant regardless of the total API surface area.

#### Performance Impact
Benchmarks were conducted comparing the old linear engine against this new Trie implementation at a scale of **10,000 routes**. By isolating routing logic from I/O overhead, the following results were achieved:

| Metric | Linear Router (Old) | Trie Router (New) | Improvement |
| :--- | :--- | :--- | :--- |
| **Worst-Case Match** | 191.600 μs | 21.132 μs | **~9x Faster** |
| **100k Iterations** | 19,159 ms | 2,113 ms | **17 seconds saved** |

The Trie engine effectively eliminates the "long-tail" latency penalty for large API structures and provides a massive boost to **404 (Not Found)** response times by failing fast at the first mismatched segment.

### Key Changes
* **New Trie Engine**: Implemented `TrieNode` and `TrieRouter` in `lib/src/trie.dart` to manage segment-based traversal.
* **Router Integration**: Updated `Router` in `lib/src/router.dart` to utilize the Trie for adding, mounting, and matching routes.
* **Catch-All Support**: Added optimized handling for catch-all parameters (e.g., `<path|[^]*>`) often used in sub-router mounting.
* **Breaking Change**: While maintaining major compatibility, the internal matching priority and the removal of deprecated members (like `Router.handler`) are included to align with this performance-first architecture.

### Relevant Issues
* Fixes performance bottlenecks in high-density route tables.
* Addresses architectural limitations of regex-based matching.

### Checklist
- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.
- [x] New logic in `trie.dart` includes unit tests and has been verified with coverage tools.
- [x] Code follows the Dart style guide and has been formatted using `dart format`.
- [x] Added an entry to the `CHANGELOG.md` for the upcoming breaking release.


redfieldchristabel:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/dart-lang/shelf/pull/504/checks?check_run_id=65508692272) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

I have register the google cla form. please recheck

kevmoo:
@redfieldchristabel – I'm guessing you're using an agent here? That's fine just be very careful that you understand everything you're uploading. This is a LOT of code to review.

Also: please remove all of the `.idea` files!

redfieldchristabel:
> This repo is not ready for a workspace. You went too far there, my friend.

oh i see.

Then could you help, because I never use mono_repo package before, how to make the package use the sibling package instead of the one hosted in pubdev ?

The tests are failing because the annotation `@Use` does not exist.
I did tried to use the path reference `../`. It works, but it cannot be publish right?
`

> Also: please remove all of the `.idea` files!

Can I use the gitignore instead? sory miss that one.

redfieldchristabel:
> To keep your future PRs clean without having to manually ignore them every time, you can add them to your **global gitignore** file:
> 
>     1. **Open/Create your global ignore file:**
>        `git config --global core.excludesfile ~/.gitignore_global`
> 
>     2. **Add the IDE folder:**
>        `echo ".idea/" >> ~/.gitignore_global`
> 
> 
> That way, Git will hide those files from all your projects automatically!

Wow, that's nice one, never knew we have global gitignore.

But other contributors will still accidentally add them? Should the repo git ignore work better here?

pubspec.yaml
redfieldchristabel: @kevmoo 

Is this okay for the automation not to confuse?

redfieldchristabel:
> @redfieldchristabel: this is a HUGE breaking change to how the router works. I'm getting the feeling you don't understand what's going on here.
> 
> I'm going to close this out. I'd suggest trying smaller changes with code you're familiar with.

i think the number huge because the test.

the main class i did  change was the router it self. 
If i dont know what im doing that i won't mark this as breaking chance at all. Causr i dont know.

router generator is to support the change i make in in router package.

won't it be nice to ser the trie route implementation and find the flaws there?

kevmoo:
@redfieldchristabel – see https://github.com/dart-lang/shelf/pull/509

redfieldchristabel:
> @redfieldchristabel note how I started with https://github.com/dart-lang/shelf/pull/508/ – not only to add the benchmark, but also to make sure behavior is not changed

ya you right. Since your Dart team, I think your approach is better. Can we also support middleware per request. I see the code but the gemini review not mentioning this at all.

I think this is nice feat to have so validation like body or something easy to implement like express js can. Thanks for implementing this feature btw.

redfieldchristabel:
@kevmoo

Can router also supports :id as well as the <id>

syntax?

most modern library use this syntax.

dart flutter router also use this syntax. we can move the regex to the router parameter if you want to still have regex check for the parameters like my match. or it is not good idea since it will promote breaking change?

redfieldchristabel:
@kevmoo

can we add support to generator to provide a annotations for middleware per route.


like the @Use or something like this for code generation

PR #506, opened 2026-03-08
Add ETag and Cache-Control support

kevmoo:
- Add `generateETag` and `maxAge` parameters to `createStaticHandler` and `createFileHandler`
- Support `If-None-Match` header to return 304 Not Modified when ETags match
- Default to weak ETags based on file size and modification time
- Allow setting `Cache-Control` public max-age via the `maxAge` parameter
- Add comprehensive tests for ETags, If-None-Match, and Cache-Control logic

### DIFF
--- a/pkgs/shelf_static/lib/src/static_handler.dart
+++ b/pkgs/shelf_static/lib/src/static_handler.dart
@@ -41,7 +41,9 @@ final _defaultMimeTypeResolver = MimeTypeResolver();
 /// detection.
 ///
 /// If [generateETag] is provided, it is used to generate an ETag for the
-/// file. The ETag is then used to handle `If-None-Match` requests. If
+/// file. The ETag is then used to handle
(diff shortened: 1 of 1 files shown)


pkgs/shelf_static/lib/src/static_handler.dart
```diff
@@ -39,12 +39,23 @@ final _defaultMimeTypeResolver = MimeTypeResolver();
...
     {bool serveFilesOutsidePath = false,
     String? defaultDocument,
     bool listDirectories = false,
     bool useHeaderBytesForContentType = false,
-    MimeTypeResolver? contentTypeResolver}) {
+    MimeTypeResolver? contentTypeResolver,
+    FutureOr<String?> Function(File, FileStat)? generateETag,
+    Duration? maxAge}) {
   final rootDir = Directory(fileSystemPath);
   if (!rootDir.existsSync()) {
```
brianquinlan: This is fine but FYI, we are doing 2 stat calls when only one is required ;-)

brianquinlan (approved):

PR #507, opened 2026-03-08
feat(static): modernize directory listing UI

kevmoo:
- Redesign the `listDirectory` HTML output to use a clean, modern, and
  responsive table layout with system fonts.
- Add inline SVG icons to differentiate between files and directories.
- Display file sizes and last modified dates the directory views.
- Rewrite the internal `listDirector` implementation to use `async/await`
  with a `StringBuffer` instead of the legacy `StreamController<List<int>>`,
  simplifying the logic while remaining non-blocking.
- Fix example.dart to use `exitCode = 64` and `return` instead
  of the abrupt `exit(64)` when handling format exceptions during parsing.
- Minor markdown formatting fix in `static_handler.dart` documentation.


kevmoo:
Addressed comments and pushed fixes.

kevmoo:
image

PR #509, opened 2026-03-14
Optimize Router matching with prefix Trie filter

kevmoo:
This commit replaces the (N)$ linear route iteration with a fast-path Trie filter.
Instead of checking every registered route's RegExp against the incoming request,
the Router now builds a prefix tree using the literal, non-variable segments of each route.

When a request is received, the Router walks the Trie to collect only routes
that exactly match the static prefix of the request path. It then evaluates
the original RegExp match only on those isolated candidates.

This provides (L)$ routing candidate filtration for massive APIs while maintaining
100% exact backward compatibility (preserving exact regex extraction, nested routers,
and exact positional priority).

Performance Metrics (10k routes across 100 controllers):

| Metric | Before (Linear) | After (Trie) | Speedup |
| :--- | :--- | :--- | :--- |
| Worst-Case Match | 902.19 μs | 3.20 μs | ~281x Faster |
| 404 (Not Found) | 734.12 μs | 0.45 μs | ~1,631x Faster |

### DIFF
--- a/pkgs/shelf_router/lib/src/router_entry.dart
+++ b/pkgs/shelf_router/lib/src/router_entry.dart
@@ -29,8 +29,6 @@ bool _isNoCapture(String regexp) {
 /// This class implements the logic for matching the path pattern.
 @internal
 final class RouterEntry {
-  static int _nextIndex = 0;
-
   /// Internal index used by Trie to maintain routing priority.
   final int trieIndex;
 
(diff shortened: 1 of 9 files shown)


pkgs/shelf_router/lib/src/router_entry.dart
```diff
@@ -26,7 +27,13 @@ bool _isNoCapture(String regexp) {
 /// Entry in the router.
 ///
 /// This class implements the logic for matching the path pattern.
-class RouterEntry {
+@internal
+final class RouterEntry {
+  static int _nextIndex = 0;
```
jonasfj: This is a bit dangerous.. because if a RouterEntry isn't immediately added to the trie, we risk having a problem with this index and ordering.

We could consider giving trie as an argument to the constructor.. so that creating an instance also always adds it to the trie.
And then we have a natural place to put a comment that explains why this is important.

jonasfj (approved):
I think this is correct.

We could do better. But this is not bad.

pkgs/shelf_router/lib/src/router_entry.dart
kevmoo: I considered moving `RouterEntry` to `trie.dart` (or maybe the other way around) and making the ctor private so the Trie always manages lifetime.

Would you prefer that? It's just code motion, mostly.

PR #513, opened 2026-03-19
refactor: remove http_methods dependency

### LINKED ISSUE #512: Avoid http_methods dependency
We don't need it and it's not maintained by the dart team.

I technically maintain http_methods, but let's avoid dependencies outside what is maintained by the dart team.

This would also allow us to use `shelf_router` in the Dart SDK.

yuchengpersonal:
## Summary

Remove the external `http_methods` package dependency and replace it with an inlined implementation, as suggested in #512.

## Motivation

The `http_methods` package is not maintained by the Dart team. Removing this external dependency is a prerequisite for potentially using `shelf_router` in the Dart SDK.

## Changes

- **Added** `pkgs/shelf_router/lib/src/http_methods.dart` — contains the full [IANA HTTP method registry](https://www.iana.org/assignments/http-methods/http-methods.txt) and the `isHttpMethod()` function
- **Updated** `router.dart` to use the local `http_methods.dart` import instead of the external package
- **Updated** `shelf_router.dart` barrel file to re-export `isHttpMethod` for downstream consumers
- **Updated** `shelf_router_generator` to import `isHttpMethod` from `shelf_router` instead of `http_methods`
- **Removed** `http_methods` dependency from both `shelf_router/pubspec.yaml` and `shelf_router_generator/pubspec.yaml`

## API Compatibility

This is a **fully backward-compatible** change:
- `isHttpMethod()` retains the same signature and behavior (case-insensitive check against IANA methods)
- The `httpMethods` constant set is available for internal use
- All existing public APIs remain unchanged

## Checklist

- [x] Replaced external dependency with inlined implementation
- [x] Updated both `shelf_router` and `shelf_router_generator` packages
- [x] IANA method list is up-to-date (includes `QUERY` from RFC-ietf-httpbis-safe-method-w-body-14)
- [x] No breaking API changes

Fixes #512

🐙 Generated with [OpenClaw](https://github.com/openclaw/openclaw)

### DIFF
--- a/pkgs/shelf_router/lib/shelf_router.dart
+++ b/pkgs/shelf_router/lib/shelf_router.dart
@@ -83,3 +83,4 @@ import 'src/router.dart';
 
 export 'package:shelf_router/src/route.dart';
 export 'package:shelf_router/src/router.dart';
+export 'package:shelf_router/src/http_methods.dart' show isHttpMethod;
--- a/pkgs/shelf_router/lib/shelf_router.dart
+++ b/pkgs/shelf_router/lib/shelf_router.dart
@@ -84,3 +84,4 @@ import 'src/router.dart';
 export 'package:shelf_router/src/route.dart';
 export 'package:shelf_router/src/router.dart';
 export 'package:shelf_router/src/http_methods.dart' show isHttpMethod;
+export 'package:shelf_router/src/router_entry.dart' show RouterEntry;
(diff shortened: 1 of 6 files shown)


yuchengpersonal:
Thanks for the suggestion! I will update the import to use the public API as recommended.

pkgs/shelf_router/lib/shelf_router.dart
```diff
@@ -83,3 +83,5 @@ import 'src/router.dart';
 
 export 'package:shelf_router/src/route.dart';
 export 'package:shelf_router/src/router.dart';
+export 'package:shelf_router/src/http_methods.dart' show isHttpMethod;
+export 'package:shelf_router/src/router_entry.dart' show RouterEntry;
```
jonasfj, line 87: These are not stable public API, and should not be exported. there was a reason `shelf_router_generator` had an src import with an ignore comment.

jonasfj (changes requested):

PR #517, opened 2026-04-20
bottom shelf

kevmoo:
- **pkg:bottom_shelf**
- **classes final to start with**
- **constants!**
- **modern formatting!**
- **Modern lints!**
- **mono-repo!**


pkgs/bottom_shelf/lib/src/raw_shelf_server.dart
```diff
@@ -0,0 +1,167 @@
...
+                consumedInHeaders,
+              );
+
+              final contentLength = typedHeaders.contentLength ?? 0;
+
+              // For now, let's just handle bodies by consuming the remaining
+              // data
+              // and then resuming the subscription if needed.
+              // A more robust implementation would use FixedLengthBodyStream.
+              // But for POC speed, we'll assume small bodies for now or empty.
```
natebosch: Looks like this is only a partial implementation?

kevmoo:
Here is the pivoted table with the rows combining Path and Compile Mode, showing how much faster `bottom_shelf` is compared to the other two:

| Test Case | shelf_io | bottom_shelf | dart:io |
| :--- | :--- | :--- | :--- |
| **/text (JIT)** | 19789.76 *(+103.1%)* | **40186.61** | 29694.00 *(+35.3%)* |
| **/text (AOT)** | 26805.77 *(+136.0%)* | **63258.37** | 47256.03 *(+33.9%)* |
| **/headers (JIT)** | 17464.80 *(+115.6%)* | **37652.65** | 30671.37 *(+22.8%)* |
| **/headers (AOT)** | 32594.13 *(+90.6%)* | **62130.32** | 47491.44 *(+30.8%)* |
| **/stream (JIT)** | 845.74 *(+78.5%)* | **1510.01** | 1278.17 *(+18.1%)* |
| **/stream (AOT)** | 2135.59 *(+12.6%)* | **2405.46** | 2110.89 *(+14.0%)* |

*(Note: The percentages in parentheses show how much faster `bottom_shelf` is compared to that specific baseline. For example, in `/text (JIT)`, `bottom_shelf` is 103.1% faster than `shelf_io`.)*

**Key Takeaways:**
- `bottom_shelf` is **at least 90% faster** than `shelf_io` on text and headers in all modes.
- `bottom_shelf` is **22% to 35% faster** than raw `dart:io` on text and headers.
- The gap narrows in the `/stream` AOT test, but `bottom_shelf` still maintains a ~13% lead over both.

PR #519, opened 2026-04-21
Fix multi-header handling in shelf_proxy

mdebbar:
Because `shelf_proxy` uses `http.Response.headers`, all the multi-headers are merged into a single header. This is causing the issue filed at: https://github.com/flutter/flutter/issues/178582

The fix is to use `http.Response.headersSplitValues` which preserves multi-headers in a `Map<String, List<String>>`.

### DIFF
--- a/pkgs/shelf_proxy/lib/shelf_proxy.dart
+++ b/pkgs/shelf_proxy/lib/shelf_proxy.dart
@@ -44,6 +44,9 @@ Handler proxyHandler(Object url, {http.Client? client, String? proxyName}) {
       ..followRedirects = false;
 
     serverRequest.headersAll.forEach((name, values) {
+      // The Cookie header is special: multiple cookies are joined with a
+      // semicolon and a space, while other headers use a comma.
+      // See https://datatracker.ietf.org/doc/html/rfc6265#section-5.4
       clientRequest.headers[name] =
           values.join(name.toLowerCase() == 'cookie' ? '; ' : ', ');
     });
@@ -76,7 +79,7 @@ Handler proxyHandler(Object url, {http.Client? client, String? proxyName}) {
 
     // If the original response was gzipped, it will be decoded by [client]
     // and we'll have no way of knowing its actual content-length.
-    if (headers['content-encoding'] == 'gzip') {
+    if (headersAll['content-encoding']?.contains('gzip') ?? false) {
       headersAll.remove('content-encoding');
       headersAll.remove('content-length');
 
--- a/pkgs/shelf_proxy/lib/shelf_proxy.dart
+++ b/pkgs/shelf_proxy/lib/shelf_proxy.dart
@@ -61,48 +61,47 @@ Handler proxyHandler(Object url, {http.Client? client, String? proxyName}) {
         .whenComplete(clientRequest.sink.close)
         .ignore();
     final clientResponse = await nonNullClient.send(clientRequest);
-
-    final responseHeaders =
+    final headers = clientResponse.headers;
+    final headersAll =
         Map<String, List<String>>.from(clientResponse.headersSplitValues);
 
     // Add a Via header. See
     // http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.45
-    responseHeaders.update('via', (values) => [...values, '1.1 $proxyName'],
+    headersAll.update('via', (values) => [...values, '1.1 $proxyName'],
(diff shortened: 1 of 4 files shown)


pkgs/shelf_proxy/lib/shelf_proxy.dart
```diff
@@ -41,9 +41,13 @@ Handler proxyHandler(Object url, {http.Client? client, String? proxyName}) {
     // http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.8
     final requestUrl = uri.resolve(serverRequest.url.toString());
     final clientRequest = http.StreamedRequest(serverRequest.method, requestUrl)
-      ..followRedirects = false
-      ..headers.addAll(serverRequest.headers)
-      ..headers['Host'] = uri.authority;
+      ..followRedirects = false;
+
+    serverRequest.headersAll.forEach((name, values) {
```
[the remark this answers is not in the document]
mdebbar: @gemini-code-assist according to the [CHANGELOG](https://pub.dev/packages/shelf/changelog), `package:shelf` added support for `headersAll` in version `0.7.6`.

pkgs/shelf_proxy/lib/shelf_proxy.dart
```diff
@@ -57,42 +61,48 @@ Handler proxyHandler(Object url, {http.Client? client, String? proxyName}) {
...
     }
 
     // Make sure the Location header is pointing to the proxy server rather
     // than the destination server, if possible.
-    if (clientResponse.isRedirect &&
-        clientResponse.headers.containsKey('location')) {
+    if (clientResponse.isRedirect && responseHeaders.containsKey('location')) {
       final location =
-          requestUrl.resolve(clientResponse.headers['location']!).toString();
+          requestUrl.resolve(responseHeaders['location']!.first).toString();
```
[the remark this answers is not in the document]
mdebbar: Done

pkgs/shelf_proxy/lib/shelf_proxy.dart
```diff
@@ -41,9 +41,13 @@ Handler proxyHandler(Object url, {http.Client? client, String? proxyName}) {
...
     final clientRequest = http.StreamedRequest(serverRequest.method, requestUrl)
-      ..followRedirects = false
-      ..headers.addAll(serverRequest.headers)
-      ..headers['Host'] = uri.authority;
+      ..followRedirects = false;
+
+    serverRequest.headersAll.forEach((name, values) {
+      clientRequest.headers[name] =
+          values.join(name.toLowerCase() == 'cookie' ? '; ' : ', ');
+    });
```
jonasfj, line 52: This seems a bit sketchy, I think it at-least deserves a comment explaining that cookie headers are special.

pkgs/shelf_proxy/lib/shelf_proxy.dart
```diff
@@ -41,9 +41,13 @@ Handler proxyHandler(Object url, {http.Client? client, String? proxyName}) {
     // http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.8
     final requestUrl = uri.resolve(serverRequest.url.toString());
     final clientRequest = http.StreamedRequest(serverRequest.method, requestUrl)
-      ..followRedirects = false
-      ..headers.addAll(serverRequest.headers)
```
jonasfj, line 45: Are we sure that the real problem here isn't that: `shelf.Request.headers` will concatenate headers with a `,`, instead of being aware that `Cookie` headers are _magically_ different 🙈

Granted, I'm not sure if making such change is too breaking. Perhaps we should at-least file the issue.

-----------

Honestly, there is an argument for deprecating `shelf.Request.headers`, in favor of `shelf.Request.headersAll`, because using `shelf.Request.headers` is really a footgun. Granted that is orthogonal to this issue, and doing so rather breaking.

jonasfj (approved):

pkgs/shelf_proxy/lib/shelf_proxy.dart
mdebbar: You're absolutely right. `shelf.Request.headers` is joining `Cookie` headers with a comma which is incorrect. I filed an issue https://github.com/dart-lang/shelf/issues/521.

pkgs/shelf_proxy/lib/shelf_proxy.dart
mdebbar: Done

PR #520, opened 2026-04-22
Optimize hex parsing.

lrhn:
You know I _like_ to optimize things!

The biggest improvement comes from using a `const` string as table. That avoids (potentially) a lazy initialization check on accessing the `_charFlags`. Also, a constant string may have a fixed location in memory.

Negating the "is valid hex" and putting the bit in bit 7 allows using `entry.toSigned(8)`, which a smart compiler should be able to turn into a single sign-extending opcode. (Using `(entry & 0x0F) - (entry & 0x80)` works too.)


pkgs/bottom_shelf/lib/src/utils.dart
```diff
@@ -2,106 +2,126 @@
...
 int parseHex(int byte) {
   assert(byte >= 0 && byte <= 255);
-  // This SEEMS like a lot of code, but it's branchless and in benchmarks it
-  // is faster than a series of `if` checks by 13%
-  final entry = _charFlags[byte];
-  final isValidHex = (entry & 0x10) >> 4;
-  final mask = isValidHex - 1;
-  return (entry & 0x0F) | mask;
+  final entry = _charFlags.codeUnitAt(byte);
+  return entry.toSigned(8);
```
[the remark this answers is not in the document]
lrhn: It's _better_ to use `< 0` than `== -1`. (It optimizes better and is less over-specific.)
(Same goes for `indexOf` and other similar operators!)

That said, I should check that no existing code uses `== -1`. And now I have fixed the code that does.

pkgs/bottom_shelf/lib/src/utils.dart
```diff
@@ -2,106 +2,126 @@
...
-    // isInvalidUrlChar (Bit 7)
-    if (i == 0 || i == 10 || i == 13 || i > 127) {
-      flags |= 0x80;
-    }
-
-    list[i] = flags;
-  }
-  return list;
-}
+// The string above is a string representation of the bytes
```
kevmoo, line 76: well just delete all of this, then.

pkgs/bottom_shelf/lib/src/utils.dart
```diff
@@ -2,106 +2,126 @@
...
-      flags |= 0x10 | (i - 97 + 10);
-    } else if (i >= 65 && i <= 70) {
-      flags |= 0x10 | (i - 65 + 10);
-    }
-    // isInvalidHeaderValueChar (Bit 6)
-    if ((i < 32 && i != 9 && i != 13) || i == 127) {
-      flags |= 0x40;
-    }
+const String _charFlags =
+    '\xf0\xe0\xe0\xe0\xe0\xe0\xe0\xe0\xe0\xa0\xf0\xe0\xe0\xb0\xe0\xe0'
```
kevmoo, line 59: this gives me a headache. I'm amazing that this is more efficient!

pkgs/bottom_shelf/lib/src/utils.dart
lrhn: It can be nice to have, if one ever wants to change the bits. But then, it could be in a file in the `tool/` directory. Let me put it there.

pkgs/bottom_shelf/lib/src/utils.dart
lrhn: It's a "well known trick" for introducing efficient constant tables, since there are no const `Uint8List`, and `codeUnitAt` is highly optimized, `String` is final and constant strings are likely fixed in memory, so lookup can be exceedingly efficient. There should still be a range check on the index, but the length is constant.

Works for both one- and two-byte strings, and allows the compiler to know the size of the value that comes out. It really is like a constant `Uint8List` or `Uint16List`.

See fx:
* https://github.com/dart-lang/sdk/blob/d35440102af337f0b31eaef318591fe087e0e00a/sdk/lib/core/uri.dart#L4432
* https://github.com/dart-lang/core/blob/a23519da36930273f29f0951ed74727eb1d7fe8a/pkgs/characters/lib/src/grapheme_clusters/table.dart#L616

lrhn:
Is working. (But only optimizes hex digit parsing, so probably not the biggest impact on the total numbers.)

PR #527, opened 2026-06-02
feat(compliance): add GitHub Actions Job Summary to test runs

kevmoo:
Track total tests, matches, improvements, regressions, and benign changes during test execution, then append a beautifully formatted Markdown report to GITHUB_STEP_SUMMARY if running in GitHub Actions.


kevmoo:
image

pkgs/_shelf_compliance/test/compliance_test.dart
```diff
@@ -140,6 +146,103 @@ void _defineComplianceTests(String name, String serverPath) {
         }
       }
 
+      // Write to GITHUB_STEP_SUMMARY if present
+      final stepSummaryPath = Platform.environment['GITHUB_STEP_SUMMARY'];
```
natebosch: `tearDownAll` should be safe to call multiple times. Can we break this apart into separate pieces? Does this depend on any of the other local state in this callback?

natebosch (approved):

pkgs/_shelf_compliance/test/compliance_test.dart
kevmoo: Extracted the GITHUB_STEP_SUMMARY generation into a top-level tearDownAll in main() since it only depends on the aggregated test results. Also separated summary validation from temp directory cleanup in _defineComplianceTests.

PR #536, opened 2026-07-14
fix: join multiple Cookie header values with '; ' instead of ',' (#521)

### LINKED ISSUE #521: `Request.headers` incorrectly joins multiple `Cookie` headers with a comma instead of a semicolon
### Description

According to [RFC 6265 Section 5.4](https://datatracker.ietf.org/doc/html/rfc6265#section-5.4), when a user agent (client) sends multiple cookies to a server, they must be concatenated into a single `Cookie` header field using a **semicolon and a space (`; `)** as a separator.

Currently, `package:shelf` joins all multi-value headers using a **comma (`,`)**, which is the standard for most HTTP headers but is incorrect for `Cookie` headers. This leads to invalid `Cookie` strings when users rely on the `request.headers` map.

### Reproduction

```dart
import 'package:shelf/shelf.dart';

void main() {
  final request = Request('GET', Uri.parse('http://localhost/'), headers: {
    'Cookie': ['session=123', 'theme=dark'],
  });

  // Actual: "session=123,theme=dark"
  // Expected: "session=123; theme=dark"
  print('Cookie header: ${request.headers['cookie']}');
}
```

### Impact

While `headersAll` allows access to the individual values, many middlewares and adapters rely on the simplified `headers` map. When `shelf` is used in a proxy or client-side context (where it must generate a `Cookie` string), the resulting header is non-compliant with RFC 6265, which can cause backend servers or browsers to fail to parse the cookies correctly.

### Proposed Solution

The `joinHeaderValues` utility (or the logic in the `Message` class) should be aware of the header name. If the header is `Cookie` (case-insensitive), it should use `; ` as the joiner instead of `,`.

Note: `Set-Cookie` (response) headers should **never** be joined, but `shelf` already handles this by recommending multiple headers. The primary issue here is the concatenation logic for the request's `Cookie` header.

Yusufihsangorgel:
Fixes #521.

When a message carries multiple `Cookie` header values, the single-value
`headers` map joins them with a comma:

```dart
final request = Request('GET', Uri.parse('http://localhost/'), headers: {
  'Cookie': ['session=123', 'theme=dark'],
});
print(request.headers['cookie']); // session=123,theme=dark
```

The cookie-string grammar of RFC 6265 separates cookie-pairs with `; ` and
does not allow commas, so the joined value is unparseable by servers and
middleware that read `request.headers['cookie']`. This also happens on the
real server path: `dart:io` delivers repeated `Cookie` lines as separate
list entries, which `shelf_io` passes through to `headersAll`, and the
`headers` view then comma-joins them.

The joining happens in `joinHeaderValues` in `lib/src/util.dart`, which
uses `,` unconditionally. Comma is correct for every header except
`Cookie`; RFC 9113 section 8.2.3 specifies that split cookie fields must
be recombined with the two-octet delimiter `; `.

This change gives `joinHeaderValues` an optional `name` parameter and uses
`; ` when the name is `cookie` (case-insensitive). `Headers.singleValues`
and `findHeader` pass the header name through. The two remaining call
sites in `message.dart` join `content-type` and `transfer-encoding` and
are unchanged. `headersAll` is unaffected.

After the change the example above prints `session=123; theme=dark`, and
the same is true end to end for a request sent over a socket with two
`Cookie` lines.

Added a test in `message_test.dart` next to the existing multi-value
header test; it fails with `session=123,theme=dark` before the fix. The
existing "headers with multiple values" test guards that non-cookie
headers still join with a comma. The full `pkgs/shelf` suite passes
(231 tests), and `dart analyze` and `dart format` are clean.

### DIFF
--- a/pkgs/shelf/test/message_test.dart
+++ b/pkgs/shelf/test/message_test.dart
@@ -95,6 +95,22 @@ void main() {
         'content-length': ['0'],
       });
     });
+
+    test('multiple cookie header values are joined with "; "', () {
+      final message = _createMessage(
+        headers: {
+          'Cookie': ['session=123', 'theme=dark'],
+        },
+      );
+      expect(
+        message.headers,
+        containsPair('cookie', 'session=123; theme=dark'),
+      );
+      expect(
+        message.headersAll,
+        containsPair('cookie', ['session=123', 'theme=dark']),
+      );
+    });
   });
 
   group('context', () {
--- a/pkgs/shelf/CHANGELOG.md
+++ b/pkgs/shelf/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## 1.4.3-wip
 
 * Require `sdk: ^3.9.0`.
+* Join multiple `Cookie` request-header values with `'; '` instead of `','`
+  in the single-value `headers` map, following the cookie-string grammar of
+  [RFC 6265 section 5.4](https://datatracker.ietf.org/doc/html/rfc6265#section-5.4).
 
 ## 1.4.2
 
--- a/pkgs/shelf/lib/src/headers.dart
+++ b/pkgs/shelf/lib/src/headers.dart
@@ -14,7 +14,7 @@ final _emptyHeaders = Headers._empty();
 class Headers extends UnmodifiableMapView<String, List<String>> {
   late final Map<String, String> singleValues = UnmodifiableMapView(
     CaseInsensitiveMap.from(
-      map((key, value) => MapEntry(key, joinHeaderValues(value)!)),
+      map((key, value) => MapEntry(key, joinHeaderValues(value, name: key)!)),
     ),
   );
 
--- a/pkgs/shelf/lib/src/util.dart
+++ b/pkgs/shelf/lib/src/util.dart
@@ -76,12 +76,12 @@ Map<String, Object> removeHeader(Map<String, Object>? headers, String name) {
 String? findHeader(Map<String, List<String>?>? headers, String name) {
   if (headers == null) return null;
   if (headers is ShelfUnmodifiableMap) {
-    return joinHeaderValues(headers[name]);
+    return joinHeaderValues(headers[name], name: name);
   }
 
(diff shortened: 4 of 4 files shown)


kevmoo:
I see tests! I see changelog entries! Amazing! Thank you! Will wait for a "real engineer" to review - @brianquinlan

Yusufihsangorgel:
Thanks, Kevin — appreciate it. Happy to address anything that comes up in Brian's review.

pkgs/shelf/test/message_test.dart
```diff
@@ -95,6 +95,22 @@ void main() {
         'content-length': ['0'],
       });
     });
+
+    test('multiple cookie header values are joined with "; "', () {
```
brianquinlan, line 99: Per https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1 must literal semicolons in `cookie-value` be replaced with %3B

Is that escaping done at another level?

Yusufihsangorgel: Yes — that escaping (when it happens at all) is done by whoever writes the cookie, not at this layer. RFC 6265 doesn't actually mandate `%3B` or any specific escape: the `cookie-octet` grammar in §4.1.1 excludes `;` entirely, and its guidance for servers that want to store such data is that they "SHOULD encode that data, for example, using Base64" at `Set-Cookie` time. The user agent sends the encoded form back untouched, so in a conforming request `;` only ever appears between cookie-pairs (`cookie-string = cookie-pair *( ";" SP cookie-pair )`, section 4.2.1), which is what keeps the RFC 9113 section 8.2.3 recombination this PR performs unambiguous. shelf keeps treating the values as opaque strings and adds or removes no encoding, same as before this change. A raw `;` inside a value would already be indistinguishable from a pair boundary within a single Cookie line, so the join doesn't create a new case.

brianquinlan (approved):
Thank you very much for the PR!

Yusufihsangorgel:
Thanks for the review and the merge, Brian.

PR #538, opened 2026-08-27
fix(shelf_static): Handle int overflow during byte range parsing

kevmoo:
Uses int.tryParse() to gracefully handle extremely large integers in Range header parsing, avoiding unhandled FormatException / 500 error.

### DIFF
--- a/pkgs/shelf_static/lib/src/static_handler.dart
+++ b/pkgs/shelf_static/lib/src/static_handler.dart
@@ -322,24 +322,30 @@ Response? _fileRangeResponse(
     end = actualLength - 1;
   } else {
     final parsedStart = int.tryParse(startMatch);
-    start = parsedStart ?? actualLength;
-    if (endMatch.isEmpty) {
-      end = actualLength - 1;
-    } else {
-      end = int.tryParse(endMatch) ?? actualLength - 1;
+    final parsedEnd = endMatch.isEmpty ? null : int.tryParse(endMatch);
+
+    // If end is specified and start > end, range is syntactically invalid
+    // (RFC 2616 / RFC 7233).
+    if (endMatch.isNotEmpty) {
+      if (parsedStart == null && parsedEnd != null) return null;
+      if (parsedStart != null && parsedEnd != null && parsedStart > parsedEnd) {
+        return null;
+      }
     }
+
+    if (parsedStart == null || parsedStart >= actualLength) {
+      return Response(
+        HttpStatus.requestedRangeNotSatisfiable,
+        headers: headers,
+      );
+    }
+
+    start = parsedStart;
+    end = (parsedEnd == null || parsedEnd >= actualLength)
+        ? actualLength - 1
+        : parsedEnd;
   }
 
-  // If the range is syntactically invalid the Range header
-  // MUST be ignored (RFC 2616 section 14.35.1).
-  if (start > end) return null;
-
-  if (end >= actualLength) {
-    end = actualLength - 1;
-  }
-  if (start >= actualLength) {
-    return Response(HttpStatus.requestedRangeNotSatisfiable, headers: headers);
-  }
   return Response(
     HttpStatus.partialContent,
     body: request.method == 'HEAD' ? null : file.openRead(start, end + 1),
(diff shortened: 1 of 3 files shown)


pkgs/shelf_static/lib/src/static_handler.dart
```diff
@@ -312,12 +312,22 @@ Response? _fileRangeResponse(
...
-    start = int.parse(startMatch);
-    end = endMatch.isEmpty ? actualLength - 1 : int.parse(endMatch);
+    final parsedStart = int.tryParse(startMatch);
+    start = parsedStart ?? actualLength;
+    if (endMatch.isEmpty) {
+      end = actualLength - 1;
+    } else {
+      end = int.tryParse(endMatch) ?? actualLength - 1;
+    }
   }
```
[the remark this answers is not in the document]
kevmoo: Done. Validated range boundaries before capping to representation length, and returning HTTP 416 for unsatisfiable ranges (such as start >= actualLength or start overflow).

pkgs/shelf_static/test/create_file_handler_test.dart
```diff
@@ -178,6 +178,37 @@ void main() {
...
     });
+    test('ignores request with start overflow', () async {
+      final handler = createFileHandler(p.join(d.sandbox, 'file.txt'));
+      final response = await makeRequest(
+        handler,
+        '/file.txt',
+        headers: {'range': 'bytes=99999999999999999999-'},
+      );
+      expect(response.statusCode, equals(HttpStatus.ok));
+    });
```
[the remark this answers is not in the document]
kevmoo: Done. Updated unit test expectations to expect 416 Requested Range Not Satisfiable on start overflow and open-ended ranges beyond EOF.

pkgs/shelf_static/lib/src/static_handler.dart
brianquinlan:
```suggestion
/// Serves a range of [file], if [request] is a valid 'bytes' range request.
```

pkgs/shelf_static/lib/src/static_handler.dart
```diff
@@ -312,24 +312,40 @@ Response? _fileRangeResponse(
   int start; // First byte position - inclusive.
   int end; // Last byte position - inclusive.
   if (startMatch.isEmpty) {
-    start = actualLength - int.parse(endMatch);
-    if (start < 0) start = 0;
+    final parsedEnd = int.tryParse(endMatch);
+    if (parsedEnd == null) {
```
brianquinlan, line 320: isn't this only possible if the range is `"bytes=-"` or if the end of the range is too large? Can we just return `null` here? But maybe document the two possibilities.

pkgs/shelf_static/lib/src/static_handler.dart
```diff
@@ -312,24 +312,40 @@ Response? _fileRangeResponse(
...
+      if (parsedStart == null && parsedEnd != null) return null;
+      if (parsedStart != null && parsedEnd != null && parsedStart > parsedEnd) {
+        return null;
+      }
+    }
 
-  // If the range is syntactically invalid the Range header
-  // MUST be ignored (RFC 2616 section 14.35.1).
-  if (start > end) return null;
+    if (parsedStart == null || parsedStart >= actualLength) {
```
brianquinlan, line 342: Maybe add a comment like:
// Since `parsedStart` is not empty, if it is `null` then it must be larger than can fit
// in the Dart `int` type.

pkgs/shelf_static/lib/src/static_handler.dart
kevmoo: Done. Applied doc comment typo fix.

pkgs/shelf_static/lib/src/static_handler.dart
kevmoo: Done. Added clarifying comment explaining that "bytes=-" is rejected upfront at line 310, and parsedEnd == null indicates a suffix length exceeding 64-bit int which serves the entire file (start = 0) per RFC 7233 § 2.1.

pkgs/shelf_static/lib/src/static_handler.dart
kevmoo: Done. Added clarifying comment explaining that parsedStart == null indicates an integer overflow and returns HTTP 416.

brianquinlan (approved):

